PR #59, opened 2021-05-27
fix(Toggle): add toggle overflow

mrleidesen:
Maybe add overflow will looks better😀

### DIFF
--- a/src/components/styled/toggle.css
+++ b/src/components/styled/toggle.css
@@ -101,6 +101,9 @@
       }
     }
   }
+  &-wrapper {
+    @apply rounded-badge overflow-hidden;
+  }
 }
 
 @keyframes togglemark-off {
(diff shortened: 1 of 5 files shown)


saadeghi:
@mrleidesen Thanks, but can you please tell me what problem does this fix?

mrleidesen:
When I click Toggle, the 'toggle-mark' will overflow border. Sorry, my english is not good, you can see this screenshot.
bug1

saadeghi:
Oh, that's the animation ([here](https://github.com/saadeghi/daisyui/blob/f9a56afb28ac306b83ce92c88fb87500cea033c6/src/components/styled/toggle.css#L106)) and it's not a bug. It's applies an elastic easing. similar to `.btn` (when you click button it gets smaller then it gets a little larger than original size, then it goes normal size again)  
Sorry but I can't merge this 😅  
If you're not happy with the animation, add `no-animation` class or change `--animation-input: .4s` CSS variable.

PR #81, opened 2021-06-14
Fixed: unchangeable height in textarea

zoghal:
hi, pouya
i find a bug in textarea component. cannot change height in textarea, because defined height and min-height.
[demo]( https://codepen.io/zoghal/pen/jOBQzvo)


saadeghi:
@zoghal thanks!

PR #107, opened 2021-07-11
Feature: add "table-active" utility

efnninja:


saadeghi:
@efnninja Thanks. This changes the background color of every `td` inside `table-active`.   
I'm not sure when this could be useful. Can you explain more please?

efnninja:
Hi !
imagine you want to select a user or post to delete and you have checkbox buttons for multi deleting 
this time  `table-active ` can help to show the selected rows

saadeghi:
@efnninja In that case we would need `tr` to have an `active` class. I will add this class today:
```html
<tr class="active">
```
I have to close this PR tho. Because it doesn't highlight one row. It will highlight **all** rows on a `<table>` with `table-active` class

efnninja:
Ok !
Thanks.

saadeghi:
@efnninja Added in version `1.9.0`

PR #147, opened 2021-08-23
Site: Allow to directly enter color hex

0x0a0d:
`<input type="color"...` doesn't allow hex color value to be pasted in, it's too complicated to use. This change allows us to make it simpler
Screen Shot 2021-08-23 at 13 18 22


saadeghi:
@0x0a0d Thanks 👍

PR #287, opened 2021-11-04
Update add-themes.vue

aeadedoyin:
Adds formatting to the CSS variable part of the doc, so it can easily be copied/used directly from the site.


saadeghi:
@aeadedoyin thanks!

PR #383, opened 2022-01-08
fix(breadcrumbs): Remove the breadcrumb separator background color

shixianqin:
As shown, the breadcrumb separator is overlaid with a global style, so I set the background of the separator to a transparent color.

image


shixianqin:
I think it's probably a style issue, but if that's the design, then close this pr.

Issue #536, opened 2022-03-02
Feature Request: Config Option for Adding Prefix to Class Names


vnphanquang:
Hi all. Firstly thanks for the amazing work. 

## Feature Request

This issue is a request to add an option in the `daisyui` config to allow customize the class names with a prefix.

## Rationale

This stems from my most recent project where I came across this issue: the words used for naming components in `daisy` is widely used. This results in two problems:

1. If `daisy` has to coexist with another styling system (in our case, we were in the process of migrating [bootstrap](https://getbootstrap.com/) to tailwind), there will be conflict.

2. Due to the generous nature of Tailwind's default [extractor](https://tailwindcss.com/docs/content-configuration#customizing-extraction-logic), if someone is working in a React `.jsx` file and has the word `modal` in some comment like the following, for example:

    ```javascript
    function Component() {
      // this is meant to be my custom modal <-- modal is here and not in class attribute
      return (<div>...</div>);
    }
    ```

    Tailwind will include the `.modal` class in the output. This results in unused and unoptimized build. Having a prefix for all components and their associated utilities will help with this.

## Proposed Implementation

Please look at the pull request linked to this issue for details. There's two parts to this.

### The Config Option

We can use [postcss-js](https://github.com/postcss/postcss-js) with the [postcss-class-prefix](https://www.npmjs.com/package/postcss-class-prefix) plugin to add prefix on demand in user's `tailwind.config.js`:

```javascript
const postcssJs = require('postcss-js');
const postcssPrefix = require('postcss-class-prefix');

//...

const prefix = config('daisyui.prefix');
if (prefix) {
  file = postcssJs.sync(postcssPrefix(prefix))(file);
}
```

### Documentation Site Improvement

In the pull request, I also added instruction about this new `option` in the config page. But more importantly, I feel like the prefix should be a user-changeable global variable (similar to the theme dropdown) that is actually reflected in the class table and the html code examples.

![docs-dynamic-prefix](https://user-images.githubusercontent.com/25895844/156366296-06c277bc-64c1-4423-a4da-cf18d36b59e6.png)

The change to html codeblocks to support dynamic prefix requires further discussion. Right now my implementation includes a [custom svelte action](https://svelte.dev/docs#template-syntax-element-directives-use-action) that replace `$$` with the `prefix` value coming out of a global store. This requires the following change to every `pre` block that wants to support this:

```html
<pre slot="html" use:replace={{ to: $prefix }}>{
`<button class="$$btn">Button</button>
[...]

PR #537, opened 2022-03-02
feat(config): allow adding prefix to class names

vnphanquang:
help with https://github.com/saadeghi/daisyui/issues/536


src/docs/src/components/Prism.svelte
```diff
@@ -7,22 +7,24 @@
...
   import "prism-svelte"
-  import { tick } from "svelte"
+  import { afterUpdate } from "svelte"
 
   export let language = "javascript"
   export let source = ""
   export let transform = (x) => x
   let element, formattedCode
 
-  $: $$props && (source || element) && highlightCode()
```
vnphanquang, line 17: replace by the `afterUpdate` below. This is to support correct re-rendering when the dynamic global prefix value changes (element won't change here, so the content is not updated accordingly)

src/docs/src/lib/actions.js
```diff
@@ -0,0 +1,21 @@
+export function replace(node, parameters = {}) {
```
vnphanquang, line 1: Example Usage:

```
<p use:replace={{ search: '$$', to: $prefix }}>
 Some content: $$ will be replaced with prefix
</p>
```

PR #587, opened 2022-03-14
fix Japanese on the landing page and sidebar

ryoppippi:
Thank you for creating DaisyUI & Japanese page!

I fixed some sentences on the landing page.

Also, I changed section names in the sidebar from Japanese terms to original English terms because Japanese developers used to this way and some words are hard to translate.


ryoppippi:
I'll fix individual section pages later.

saadeghi:
Thank you 🙌

PR #622, opened 2022-03-16
Parameterise modal background colours

JamesYeoman:
Since my I've designed my app to use the darker colours at the back and the lighter colour at the front, it felt weird that the modals didn't follow the same colour scheme.

It would make sense to have this override behaviour, considering it's available for the lifted tabs (although surprisingly not available for the boxed tabs).


saadeghi:
This is not required at all since we can use any custom color for `modal` or `modal-box`:  
Example: https://play.tailwindcss.com/oBdSOJHq55

PR #782, opened 2022-05-09
Update Language `Select` component title

rizzrark:
The title in the language `Select` component is set to _Change Theme_ instead of _Change Language_


saadeghi:
@rizzrark Thank you!

PR #786, opened 2022-05-11
Add Traditional Chinese (Taiwan) translation

SiderealArt:
I've also

-  increased the width of the language dropdown so that `Traditional Chinese (Taiwan)` name fits on one line.

- fix the html tag in docs/themes

image


saadeghi:
@SiderealArt Thank you!

PR #813, opened 2022-05-22
Make last sample on /components/swap actually swap

edman:
Add a hidden checkbox to the last sample to make it work.

(disclaimer: I'm making this PR from the github editor and I haven't actually tried this code, but judging from other code around it seems this is what needs to be done)


saadeghi:
That example supposed to show how the class name `swap-active` controls the visibility. It doesn't need to have a checkbox. It's for a situation where people need to show/hide the content by adding `swap-active` class.

PR #858, opened 2022-06-13
Update radio.svelte.md: fix Checkbox->Radio in first example

outluch:
Replaced Checkbox to Radio in first example


saadeghi:
Thank you

Issue #860, opened 2022-06-14
Doc improvement: Open the sidebar on Meta + K


CapitaineToinon:
# What's the issue

I often work with 2 browsers open in split screen, leaving only half of the space available. When I open the documentation, the sidebar is therefore collapsed and when I hit Meta + K to search on the documentation, nothing happens.

# What was I expecting?

I was expecting the sidebar to open with the search input focused. 

# How can it be fixed

After looking through the code, it seems that the solution would be pretty simple. In `src/routes/__layout.svelte`, add the same event listener than the one already used in `Search.svelte` to open the check the checkbox responsible for opening the sidebar on. I think it would also be a good idea to add an even dispatcher to the Search component so the sidebar could also automatically close on search. 

I've forked the projected and looked into solutions, I would be happy to make a pull request.

CapitaineToinon:
Done in #861 and #870 !

PR #861, opened 2022-06-14
[documentation site] Opening the sidebar on Meta + K

CapitaineToinon:
Fix for #860. I saw on contribution guide that I could open a pull request directly for modifications on the documentation website. Let me know if that's okay. 

These changes simply open the sidebar by checking the relevant checkbox whenever Meta + K is used. I also added an even listener to automatically close the sidebar whenever a search result is picked. Because the sidebar is always opened on desktop by using css, these changes only affect the small screen views.

Let me know if I should change anything.
Love daisyui but this issue was driving me crazy 👍

### DIFF
--- a/src/docs/src/components/Search.svelte
+++ b/src/docs/src/components/Search.svelte
@@ -27,12 +27,13 @@
     if ((e.keyCode === 75 && e.metaKey) || (e.keyCode === 75 && e.ctrlKey)) {
       e.preventDefault()
       seachboxEl.querySelector("input[type=search]").focus()
+      dispatch("focus")
     }
   }
 
   function onSelect({ detail }) {
     goto(searchIndex[detail.originalIndex].href)
-    dispatch('search', detail)
+    dispatch("search", detail)
   }
 </script>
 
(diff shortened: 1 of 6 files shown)


CapitaineToinon:
Switched to using media query in Javascript to conditionally show the search input in the sidebar instead. That way, the sidebar won't listen to the Search events when the Search input isn't in the DOM anymore. Now as far as I can tell it works as expected. 

The source code of useMediaQuery and useBreakpoints are basically a direct copy of @vueuse's [useMediaQuery](https://vueuse.org/core/usemediaquery/) and [useBreakpoints](https://vueuse.org/core/usebreakpoints/). 

The only concern left I have is that according to tailwind's doc, importing the entire tailwind config in js will increase bundle size and that we should use something like [babel-plugin-preval](https://github.com/kentcdodds/babel-plugin-preval) to statically include the config we need in the code.

saadeghi:
Hey.  
Thank you for the PR, but I'm not sure if the hotkeys are even necessary for the small screen.

CapitaineToinon:
For mobile phones I agree but right now the search bar disappears at 1024px, which is quite big still. In my use case on my laptop, just having 2 browsers in split screens already hit that breakpoint.

saadeghi:
Thank you 🙌

Issue #221, opened 2021-10-01
Suggestion: SnackBar component


kaname-png:
I see daisyUI doesn't have snackbar component, it would be great if it did.

saadeghi:
@kaname-png Yes, this is in my to-do list but with "toast" name.  
I believe a "toast" and a "snackbar" is the same thing. Right?

kaname-png:
@saadeghi Yes, they are the same. I hope you can implement it!

priyang12:
hi, can I work on this????

saadeghi:
@priyang12 sure. let me know if I can help

priyang12:
@saadeghi hi I have cloned the repo and installed the packages and went to src/docs and installed the packages and run the development by yarn run dev. it is giving some import error

<img width="1225" alt="Screenshot 2022-01-28 at 10 20 13 PM" src="https://user-images.githubusercontent.com/72823974/151588034-4e74b481-3554-49da-b44f-fc31b7b24b94.png">

do I need to change some imports in components postcss config file ???
<img width="476" alt="Screenshot 2022-01-28 at 10 21 58 PM" src="https://user-images.githubusercontent.com/72823974/151588318-5e83e6ba-9906-4653-9ae0-605165d4ba1a.png">

saadeghi:
@priyang12 make sure you build the daisyUI node package first: 
 https://github.com/saadeghi/daisyui/blob/master/.github/CONTRIBUTING.md#building-on-local

cryptodeal:
Recently switched to using DaisyUI with SvelteKit and was looking to replace a svelte toast library with a custom store/DaisyUI implementation of toast notifications. I've forked the repository and built a working component consistent with your vision of the toast component & accompanying wrapper by largely imitating the styles of the Alert component.

If you can assign this to me, I'll gladly submit a PR with the changes as I'd likely need help with the animations, to clean up styles. 

https://user-images.githubusercontent.com/39316528/174497938-8669ee80-c9ca-4886-b9f5-dfd08ccc3e8f.mp4

saadeghi:
We can close this issue now that we have [`toast` component](https://daisyui.com/components/toast/) (added in daisyUI `2.19.0`) thanks to @cryptodeal

PR #909, opened 2022-06-28
add basic implementation of toast component/utilities

cryptodeal:
Here's my super basic implementation of a toast component.

Would love feedback on the styles and and any help I can get on adding css animations as I'm just leveraging svelte/transitions when using my my other svelte projects at the moment. 

Additionally, I've added the corresponding routes/components/toast.svelte.md (published set to false), a Toast component, and a notification store (located in $lib/stores) to make a working click to add toast notification demo in the component doc page.


ctwhome:
I love the fact that is it a full example for Svelte, but I don't think that it fits within Daisy UI since it is a CSS framework.
IT would be lovely to have a Svelte + DaisyUI lib components

cryptodeal:
> I love the fact that is it a full example for Svelte, but I don't think that it fits within Daisy UI since it is a CSS framework.
> IT would be lovely to have a Svelte + DaisyUI lib components

I'd written the component per the architecture described by @saadeghi in the original Issue/Feature Request for snackbar/toast, "Yes, it should be something similar to an alert with position:fixed but I'm going to also add animations and a wrapper to stack multiple items on top of each other. Then all we need is a template loop (with Alpine, Vue, React, etc) to add/remove items in the wrapper." See: https://github.com/saadeghi/daisyui/issues/221#issuecomment-941270950

With that being said, if that isn't the route ya'll want to go, I can just remove the Toast component and associated notifications store, which would remove any reliance on svelte by removing the dynamic click to add toast demo.

cryptodeal:
@saadeghi probably need feedback with regards to your thoughts on the current implementation of styles for the core toast component as well as the toast-control wrapper styles in order to make further updates and ensure I'm approaching this in a manner consistent with the design principles of the library.

Additionally, I'm happy to remove the svelte component and associated store I'd written to demo the template loop to add/remove toast items on click as it's reliant on a framework vs pure css.

saadeghi:
Thanks for the PR.  
I will review it soon, but I think we should avoid framework-specific JS for the components.  

I appreciate your efforts for creating the functionality of adding/removing toasts using Svelte store but I think we should just provide the style and structure for daisyUI components.  
Adding and removing elements must be handled by the developers based on their needs (they might need permanent toast, timers, close buttons, etc.)

cryptodeal:
No worries; I'd just copied the svelte implementation from another project, so didn't involve any added work on my part! I've updated the PR to remove the svelte implementation from the docs site.

saadeghi:
Added in `2.19.0` version 🎉
Thank you @cryptodeal @ctwhome    

I applied some changes tho:  
This PR was suggesting very similar styles to `alert` component. Instead of that, I used `toast` class as the wrapper and it can contain `alert`, `card` or anything else. It just handles the positioning, stacking and animation.

PR #915, opened 2022-06-30
Update drawer from hardcoded media query to tailwind screen size

seltar:


saadeghi:
Hey, thanks for the PR.  
But I'm not sure if `@media screen(lg)` is going to work 🤔

PR #918, opened 2022-07-02
issue#785 & feat: added pinned-row, pinned-col modifiers to table styled component alongwith examples in the docs

imtheaman:
table

using these modifiers user can create an scrollable table component which will cover up the dimensions of the container.
in the example i've used a fixed height of ```h-96``` and width of perhaps ```w-4xl```.

## Responsiveness
if width of the container is fixed, responsiveness of entire component depends on the responsiveness of the container containing the table component.
if width isn't fixed then, it's responsive.
Thanks.


src/experiments/playground/src/pages/default.astro
```diff
@@ -1,7 +1,660 @@
...
-
-<BaseLayout>
-  <button class="btn">Hello daisyUI</button>
-</BaseLayout>
\ No newline at end of file
+<html>
+  <head>
+    <title>Playground</title>
+  </head>
+  <body class="flex flex-col justify-center items-center min-h-[100vh]">
```
NikarashiHatsu, line 8: It works, but why don't you use

```html
<BaseLayout>
  <div class="flex flex-col justify-center items-center min-h-[100vh]">
    <!-- content -->
  </div>
</BaseLayout>
```

Just curious if building HTML skeleton from the ground up will make any difference.

.gitignore
```diff
@@ -1,5 +1,5 @@
-node_modules
-yarn.lock
+**node_modules
+**yarn.lock
```
NikarashiHatsu, line 2: Does `**` really matter?

I mean, it won't detect any `node_modules` or `yarn.lock` in the other subdirectories such as `src/docs` because it has its own `.gitignore`, right?.

saadeghi:
Thank you for the PR and inspiration. I added this to the new version `3.0` in a different way. There was so much more that I had to change.

Thanks.

PR #921, opened 2022-07-03
docs(bottom-navigation): :memo: Update bottom navigation readme

imanmalekian31:
Fix with custom colors example


saadeghi:
Thank you

PR #929, opened 2022-07-06
fix: #928

### LINKED ISSUE #928: Border in .prose pre code overlaps content
You can see the issue here:
https://play.tailwindcss.com/ZETRuWiGyw

The problem is that [`border-radius` is being set to `none`](https://github.com/saadeghi/daisyui/blob/master/src/components/styled/typography.css#L30), which [is not a valid value for that property](https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#values). It should be set to `0`.

This is how it looks with the proposed fix:
https://play.tailwindcss.com/Yh9Oe7SI77

jprinaldi:
Fix #928 

`none` is not a valid value for the `border-radius` CSS property.

Reference: https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#values

### DIFF
--- a/src/components/styled/typography.css
+++ b/src/components/styled/typography.css
@@ -27,7 +27,7 @@
   }
   pre{
     code{
-      border-radius: none;
+      border-radius: 0;
       padding: 0;
     }
   }


saadeghi:
Thank you

PR #949, opened 2022-07-15
Feat/comment

imanmalekian31:
New Comment component.

### DIFF
--- a/src/components/unstyled/comment.css
+++ b/src/components/unstyled/comment.css
@@ -1,0 +1,15 @@
+.comment {
+  @apply flex items-center py-2 px-4 font-light;
+
+  &-message {
+    @apply flex flex-col pl-2;
+  }
+
+  &-info {
+    @apply flex gap-2;
+  }
+
+  &-actions {
+    @apply flex items-baseline gap-2;
+  }
+}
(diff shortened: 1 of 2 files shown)


src/components/unstyled/comment.css
```diff
@@ -2,7 +2,7 @@
   @apply flex items-center py-2 px-4 font-light;
 
   &-message {
-    @apply flex flex-col pl-2;
+    @apply flex flex-col;
```
PadowYT2, line 5: This was a bad change, without it, the image and the text are linked

PadowYT2 (changes requested):

taylorjdawson:
@imanmalekian31 @PadowYT2 what is the status on this?

imanmalekian31:
@taylorjdawson, @saadeghi should answer it :)

PadowYT2:
And I'm not a dev, just a user

saadeghi:
It's on hold. We need to be careful.  
I know it's exciting to get new components faster but of we don't consider all use cases for a component, we will have unnecessary breaking changes, forcing everyone to change their HTML, and that's not good at all.  
I hope you understand and give it time because reviewing PRs like this is not just reviewing the code. It's about design and testing it will all possible common use cases a designer may think of for that specific component.

saadeghi:
Moving forward with version `3.0` I have no plans for adding a `comment` class for now. There are many designs that can be implemented and we need to be flexible with that however the main difference between different designs of a comment box is mostly the grid/flex layout which is already easy to make using Tailwind utility classes.

Thanks for the PR

PR #961, opened 2022-07-19
Updated example code to match preview

barnwell:
Updated example code to match what is shown in the preview


saadeghi:
Thank you

PR #969, opened 2022-07-22
Allow complementing the pseudo-classes that control the state (e.g. :checked) of a component with any selector of choice.

StoraH:
https://github.com/saadeghi/daisyui/issues/970


StoraH:
https://github.com/saadeghi/daisyui/issues/970

saadeghi:
`[aria-checked=true]` selectors are added in `2.20.0` which fixes https://github.com/saadeghi/daisyui/issues/970  
So this PR is not needed anymore.

PR #986, opened 2022-07-28
Modify translation style

classbinu:
Edited for unity of translation.
'will be' in 234 is translated as '될 것입니다', but 'will be' in 235 is translated as '됩니다'. 
Most of the 'will be' on the current translation page has been translated as '될 것입니다'.


saadeghi:
Thank you!

PR #1002, opened 2022-08-05
fix: grammatical errors

WickedInvi:
Fixed grammatical errors I found in the English documentation.


saadeghi:
Thank you  
But the JSON key names are being used inside markdown files and they should be the same key name on all other translation JSON files.  

So if you're going to change the key names on the JSON, you should find+replace the key name on all files (.svelte markdown files and translation JSONs)

WickedInvi:
Thank you for the feedback. I will go ahead and replace all the other keys too and create a new pull request when done.

Issue #875, opened 2022-06-19
Unwanted scroll while inputing to search box


tamagoez:
When I focus to search box(click or `CTRL+k`) and input something, it automatically scroll.
I found this issue on Chrome(Windows).
Thank you.

mvz:
I have the same problem in Firefox on Linux. Also, cursor keys and page up/down don't work outside the search box. I can only scroll using the touch pad.

CapitaineToinon:
The unwanted scroll seems to be caused by bad css on the daisyui doc and can be fixed. However the arrow/page up keys not working is caused by [sveltekit setting up a tabindex="-1" on the body tag](https://github.com/sveltejs/kit/issues/3501). This issue was resolved in later versions but when I tried upgrading daisyui's doc site to the latest version of sveltekit, I had a bunch of css issues.

saadeghi:
I'm afraid removing `scroll-padding-top` is not worth it. ([reason](https://github.com/saadeghi/daisyui/pull/1008#issuecomment-1209627061))

mvz:
I would guess the unwanted scrolling is javascript related, so removing `scroll-padding-top` might stop the actual effect from happening, the cause is probably still there.

CapitaineToinon:
Ah yeah fair enough about `scroll-padding-top` but I don't believe it's javascript related. As far as I understand, the input is considered outside of the view because of the scroll padding and therefore makes the browser try to bring it in view whenever the input is focused or typed in. You can even see the jumps increase if you also increase the `scroll-padding-top` property. I tried replacing the `svelte-typehead` component by a simple `input` tag, disabled javascript and the problem still persist.

saadeghi:
@CapitaineToinon 
I agree.  
I don't have a proper solution for this right now. I will try to see if this  approach fixes the issue:  
Removing the `scroll-padding-top` while typing inside the search box and then add it again after search box loses focus 😬

mvz:
I was able to reproduce the scrolling problem with pure css:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
<style>
body {
  margin: 0;
}

.content {
  max-width: 40em;
  margin-left: 16px;
  margin-right: auto;

}
.sticky-header {
  position: sticky;
  top: 0;
  padding: 16px;
  background: white;
}
.drawer {
  overflow: hidden;
  height: 100dvh;
}
.drawer-content {
  scroll-behavior: smooth;
  scroll-padding-top: 5rem;
[...]

PR #1008, opened 2022-08-09
[doc ]Fixed navbar unwanted scroll behavior

CapitaineToinon:
Fixes for #875.

Like stated in the issue, the `tabindex` problem has been fixed in the latest version of svelte-kit but attempting the upgrade introduced other issues, such as css selector changing specificities, breaking some global styles.

When in comes to the unwanted scroll behavior, removing the `scroll-padding-top` property from the layout seems to do the trick. Even after reading and testing how this property behaves (never used to before to be honest!), I don't see how it was impactful or what the intended behavior was. Let me know if this PR breaks anything.

### DIFF
--- a/src/docs/package-lock.json
+++ b/src/docs/package-lock.json
@@ -6,7 +6,7 @@
     "": {
       "dependencies": {
         "color": "^4.2.1",
-        "daisyui": "^2.11.1",
+        "daisyui": "^2.22.0",
         "emoji-unicode": "^2.0.1",
         "lodash.debounce": "^4.0.8",
         "nanoid": "^3.3.1",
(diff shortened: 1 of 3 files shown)


saadeghi:
`scroll-padding-top: 5rem` causes the page scroll to have a top padding when we open an anchor link on the page.  

For example this link:  
https://daisyui.com/components/alert/#success-color  
points to a specific section of page (the element with `#success-color` ID).  

**Without `scroll-padding-top` page scrolls to to element (without considering the fixed navbar):**
>without

**With `scroll-padding-top` page scrolls to to element with extra padding on top:**  
>with

So I don't think removing `scroll-padding-top` would be a trade-off, causing another UX issue.

PR #1026, opened 2022-08-14
Add 🇷🇺 translations

pheianox:
Fully translated to Russian 

So that people from all around world can use this beautiful plugin


saadeghi:
Thank you

PR #1028, opened 2022-08-15
Fixed typo

WickedInvi:
I've amended the files and removed the other changes.

Note after I set up prettier to require a config file it still did the unnecessary changes. I had to delete the contents of prettierrc to be able to save the file with the correct changes.


saadeghi:
Thank you

PR #1039, opened 2022-08-20
feat(docs) save to localStorage when prefix input

kagurazaka-0:
@saadeghi hello!
I prefix `ds-` when using daisyui.
I had to type `ds-` every time I looked at the docs.
It's a hassle, so I implemented it to save the prefix entered in the docs.
Thank you for creating a cute UI library!

- install [svelte-local-storage-store](https://github.com/joshnuss/svelte-local-storage-store)
- update svelte(3.46.4→^3.49.0)
  - svelte-local-storage-store required `"svelte": "^3.48.0"`. 
- feat save to localStorage when prefix input.


saadeghi:
Thank you for the effort but... I'm not sure if this is the right thing to do.  
Because it will cause confusion. People may put a value for the prefix input and the value applies to all class names on all pages as long as the page is not closed. However, user never clicked on a save button to set that value as a custom preference for the site so it makes sense when that preference clears after a page refresh.  
People may add a prefix value just to try things and they may open daisyUI website again, one month later for a completely different project. If the prefix is still there, that will cause confusions.

kagurazaka-0:
OK... but I want to store prefix in localStorage.

PR #1085, opened 2022-09-01
New translation (Portuguese)

nmferraz:
Portuguese is one of the most spoken languages on the planet, so there are also a huge number of programmers who speak this language. So it's important to have the largest number of documentation in Portuguese, and that's exactly what I did.


saadeghi:
Thank you my friend

PR #1098, opened 2022-09-05
70% smaller mask size

kidonng:
18495 bytes -> 5540 bytes ([source file](https://github.com/saadeghi/daisyui/blob/77506ca8e188a76aa318c2371e91ee1f7c18dacb/src/components/styled/mask.css), nested rules not unwrapped)

```js
import fs from 'fs'
import {optimize} from 'svgo'
import toURI from 'mini-svg-data-uri'

const original = fs.readFileSync('mask.css', 'utf8').split('\n')
const optimized = original.map((line, index) => {
    if (!line.includes('url')) return line

    const [match, mask] = /url\(.+;base64,(.+)\)/.exec(line)
    const {data} = optimize(atob(mask))
    return line.replace(match, `url("${toURI(data)}")`)
}).join('\n')

fs.writeFileSync('mask.optimized.css', optimized)
```


saadeghi:
Great job! thank you 👏

PR #1125, opened 2022-09-12
Fix: strokeLineCap in navbar example

mikerhyssmith:
Fix missspelling of strokeLinecap in Navbar example in the docs


saadeghi:
Thank you!

PR #1141, opened 2022-09-15
Docs: Update prefix reference link in ClassTable

naffydharni006:
Prefix reference link was changed from `/docs/config/` to `/docs/config/#prefix`


saadeghi:
Thank you

PR #1159, opened 2022-09-19
Fix border radius on mixed type buttons

danawoodman:
Currently, if you have buttons of different types (`a` vs `button` for example), the button styling breaks because the selectors depend on the button type:

Screen Shot 2022-09-19 at 1 34 22 PM

Change to using child selectors as this doesn't care what type of child exists.

Note I haven't directly tested this change against DaisyUI's demos, etc, instead I applied these styles locally to verify:

```css
.btn-group .btn {
	@apply rounded-none;

	&:first-child {
		@apply rounded-l-lg;
	}
	&:last-child {
		@apply rounded-r-lg;
	}
}
```

PS any reason not to use `@apply` in the code?

### DIFF
--- a/src/utilities/styled/button.css
+++ b/src/utilities/styled/button.css
@@ -1,20 +1,20 @@
 /* group */
 .btn-group,
 .btn-group.btn-group-horizontal {
-  .btn {
+  .btn:not(:first-child):not(:last-child) {
     border-top-left-radius: 0;
     border-top-right-radius: 0;
     border-bottom-left-radius: 0;
     border-bottom-right-radius: 0;
   }
-  .btn:first-child {
+  .btn:first-child:not(:last-child) {
     @apply -ml-px -mt-0;
     border-top-left-radius: var(--rounded-btn, 0.5rem);
     border-top-right-radius: 0;
     border-bottom-left-radius: var(--rounded-btn, 0.5rem);
     border-bottom-right-radius: 0;
   }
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you.

note: I added `:not()` selectors again because without them, border radius would be wrong.

danawoodman:
Sweet!

Curious tho, why would what I posted not work with border radius? From my testing it worked fine with one or many buttons in a button group

saadeghi:
This is long and confusing to explain 😅 The problem is with the responsive variations.  
We need `top-left` and `top-right` radius for the first button on the vertical group but we need to reset them to `0` when we change it to horizontal on another resolution. So when when it's horizontal, we need to set radius for `top-left` and `bottom-left` and we also need to reset the `top-right` and `bottom-right` to `0`. Because of that, the first button on a vertical group must have `bottom-right` and `bottom-left` of `0` radius but when it is also the last button (single button on the group), it will also have `top-right` and `top-left` of `0`. That will the last rule of radius to override the other one so the single button of a group will not have radius at the start because `:last-child` comes after `:first-child` and it will set the starting radius as `0`.

This is the current style: https://play.tailwindcss.com/a7e75d4XUZ?file=css  
This is how it will look like without `:not()` selectors: https://play.tailwindcss.com/RW40QC2B33?file=css

danawoodman:
Nice, thanks for the description!

Issue #1069, opened 2022-08-29
Consider removing `.text-2xs` utility


kidonng:
https://github.com/saadeghi/daisyui/blob/bb7e3f523b3be1fc73373b0f0ce6577f5eebd8b5/src/utilities/global/fontSize.css#L1-L3

- `.text-xs` in Tailwind is already `0.75rem`, [making them almost indistinguishable](https://play.tailwindcss.com/umVUkLadSm)
- [No other reference in the repo](https://cs.github.com/saadeghi/daisyui?q=text-2xs)
- [The only mention](https://github.com//saadeghi/daisyui/issues/50#issuecomment-1150209930) of `.text-2xs` in this repo noted it caused glitching

PR #1181, opened 2022-09-26
chore: add fontSize(text-2xs,text-xxs)  

kirklin:
The XS in Tailwind has a size of 0.75 REM. Since 2XS is larger than XS, the value should be greater than 0.75 REM. Therefore, XXS is more appropriate.

https://tailwindcss.com/docs/font-size
#1069

### DIFF
--- a/src/utilities/global/fontSize.css
+++ b/src/utilities/global/fontSize.css
@@ -1,3 +1,6 @@
+.text-xxs {
+  font-size: 0.6875rem;
+}
 .text-2xs {
-  font-size: 0.7rem;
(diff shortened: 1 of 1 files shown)


saadeghi:
`xs` text size having the same name with Tailwinds`s `xs` but with a different value is indeed a problem but I think the solution would be removing it and only use the size Tailwind already has. Adding another size like `xxs` etc. wouldn't fix anything. 

I can't merge this PR. I just need to review every element using the `xs` size to make sure everything looks good with the other size, then I will delete the duplicate `xs` style.

PR #1232, opened 2022-10-10
fix: #1218 - Add missing modifiers on dropdown and tooltip

### LINKED ISSUE #1218: Dropdown and Tooltip missing one direction
Hello guys, I have a use case that is not possible to achieve right now.

The dropdown component has `top`, `left`, and `right` directions. The `bottom` is the default one. It works for most use cases but fails when I need the dropdown to appear on top on mobile and bottom on desktop. That's because once I add `dropdown-top`, there is no class to override the `top` style, like `lg:dropdown-bottom`. According to the tailwind docs, [there is no way to target a single breakpoint.](https://tailwindcss.com/docs/responsive-design#targeting-a-single-breakpoint)

The same happens with the Tooltip, but the `top` direction is missing.

If it makes sense, I can open a PR with the change.

brunocalou:
This PR adds the `dropdown-bottom` and `tooltip-top` modifiers. They are needed for the following use case:

* Show dropdown on top on mobile and on bottom on desktop. It's impossible without the `dropdown-bottom` modifier, since the `dropdown-top` would be applied for all screen sizes and there would be no class to override it. e.g.

```html
<div class="dropdown dropdown-top lg:dropdown-bottom">
```

The same use case applies to the tooltip, but the bottom and top are inverted. e.g.
```html
<div class="tooltip tooltip-bottom lg:tooltip-top">
```

## Video
https://user-images.githubusercontent.com/5948318/194936865-8fabeff4-90da-4fbf-88c1-d18822af7341.mov

### DIFF
--- a/src/components/styled/dropdown.css
+++ b/src/components/styled/dropdown.css
@@ -1,6 +1,9 @@
 .dropdown .dropdown-content {
   @apply origin-top scale-95 transform transition duration-200 ease-in-out;
 }
+.dropdown-bottom .dropdown-content {
+  @apply origin-top;
+}
 .dropdown-top .dropdown-content {
   @apply origin-bottom;
 }
--- a/src/components/styled/tooltip.css
+++ b/src/components/styled/tooltip.css
@@ -33,7 +33,7 @@
   display: block;
   position: absolute;
 }
-.tooltip {
+.tooltip, .tooltip-top {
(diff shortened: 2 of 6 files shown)


saadeghi:
Thank you 🙌  
I will check the changes today and I will merge it if everything's okay.

PR #1241, opened 2022-10-15
Update stat doc to mention `stat-actions`

lucasra1:
`stat-actions` was missing in the table, but used in an example below


saadeghi:
Thank you!

PR #1274, opened 2022-10-23
Add neutral color variants

NayamAmarshe:
Hi, this PR adds neutral variants just like base.
So base has base-100, base-200 and base-300 but I felt that having neutral would benefit DaisyUI a lot since I couldn't find a better way for text coloring.

So, this is a PR that introduces that feature with:
```
neutral
neutral-200
neutral-300
```
This won't break anything as neutral is treated as neutral-100. I'm sorry for creating a PR directly without creating an issue but I thought it'd be better this way. Please let me know if there's anything wrong in my PR and if the idea of adding neutral variants is not a good one. I wouldn't mind any of your decisions, thank you.

### DIFF
--- a/src/colors/functions.js
+++ b/src/colors/functions.js
@@ -1,82 +1,101 @@
-const Color = require("color");
-const colorNames = require("./colorNames");
+const Color = require("color")
+const colorNames = require("./colorNames")
 
 module.exports = {
-
   generateForegroundColorFrom: function (input, percentage = 0.8) {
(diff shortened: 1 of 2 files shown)


saadeghi:
Can you please save your file without formatting?  
Because this PR is really really hard to review when your editor changed formatting on all the lines where there's no edit (adding `;`s, indentations, etc.)  
Only the lines with actual changes should be committed.

PR #1278, opened 2022-10-25
docs: Error from comments in modal react examples

princecodes247:
Replace unsupported HTML comments with react comments in the modal page


saadeghi:
Thank you!

PR #1352, opened 2022-11-13
fix: remove collapse-content cutoff

bvanderdrift:
In a personal project I have a collapsed list that exceeds `9000px` when opened. As far as I understand it this value is a probably-big-enough value to offset the `max-height: 0` when the collapse is closed. Since the CSS selector is more specific `max-height: none` in the `collapse-content` should also create the same behaviour while allowing for any height.

Let me know if I'm missing something.


saadeghi:
I'm not sure... But as far as I know, CSS transition animation won't work for `none` max height.

PR #1354, opened 2022-11-13
🐛 fix(bottom-navigation): Bottom navigation respects the iOS safe-area

bastiaanv:
Currently, the bottom navigation does not respects the iOS safari safe area. Implemented from: https://webkit.org/blog/7929/designing-websites-for-iphone-x/


saadeghi:
Thank you!

PR #1384, opened 2022-11-23
Fix visible scrollbar in drawer-end animation

phisn:
`overflow-hidden` on the `drawer-side` removes the scrollbar while animating in the `drawer-end` example.


saadeghi:
I see the issue.  
Thanks for the PR but this should be fixed in the component itself, not the example.  
I will work on it.

Issue #1426, opened 2022-12-11
Fix: The Collapse component loses focus and collapses when nested elements get focused


cslcdr:
The collapse classes should rely on focus-within instead of focus

PR #1428, opened 2022-12-12
fix: #1426 collapse uses focus-within instead of focus

cslcdr:
I made the change, it works well.


saadeghi:
Thank you!

saadeghi:
I had to revert this PR because it was breaking the "collapse with checkbox" functionality. It wouldn't close on second click.

I will investigate more tomorrow to see what's wrong.

PR #1459, opened 2022-12-25
update japanese translation

kumitatepazuru:
Some parts were not translated into Japanese, so they were added.


saadeghi:
Thank you

PR #1486, opened 2023-01-05
:sparkles: added colors of apple(exit,minimize and maximize)

prabincankod:


prabincankod:
Screen Shot 2023-01-05 at 5 49 22 PM
>which should look something like this

saadeghi:
Thank you but I think the current look is better.  
The bullets are just there because they are the most iconic way for that rectangle to look like an OS window. I don't want it to look exactly like apple. They are not even real buttons...  
But if you want it with those colors, you can always customize it on your CSS file or `tailwind.config.js` file. For a specific theme or for all themes.

PR #1493, opened 2023-01-07
Update Search component in docs to show the search icon

rich-info:
It is more intuitive for the user to immediately recognize the search box. That is why the icon is always displayed.


saadeghi:
Thank you

PR #1505, opened 2023-01-09
fix(docs): Bordered menu item preview has incorrect class

koolskateguy89:
As you can see in https://daisyui.com/components/menu/#item-with-border, the item that is meant to have a border, does not.


saadeghi:
Thank you!

Issue #1508, opened 2023-01-09
Add pointer-events-none by default to .collapse:focus class


kimdre:
To close a [collapse container](https://daisyui.com/components/collapse/) you need to click outside of it to remove the focus. Imho closing it should also be possible with a second click inside the container.
This is at the moment only possible by using the [collapse with a checkbox](https://daisyui.com/components/collapse/#collapse-with-checkbox).

As a workaround you can add this to your custom .css file:

```css
.collapse:focus {
    @apply pointer-events-none;
}
```

Can we perhaps have this feature implemented by default for non-checkbox collapses?

saadeghi:
Regarding the conversation on the PR, I couldn't find a CSS solution for this to get closed by both losing focus or second click.  
The only solution for this is writing some JS to uncheck the checkbox when it's clicked outside.

PR #1509, opened 2023-01-09
#1508 Make non-checkbox collapse closable on second click

kimdre:
To close a [collapse container](https://daisyui.com/components/collapse/) you need to click outside of it to remove the focus. Imho closing it should also be possible with a second click inside the container.
This is at the moment only possible by using the [collapse with a checkbox](https://daisyui.com/components/collapse/#collapse-with-checkbox).

This implements this workaround/possibility as described in #1508

### DIFF
--- a/src/components/styled/collapse.css
+++ b/src/components/styled/collapse.css
@@ -27,6 +27,11 @@
 .collapse:focus:not(.collapse-open):not(.collapse-close) .collapse-title {
   cursor: unset;
 }
+
+.collapse:focus {
+    @apply pointer-events-none;
+}
+
 .collapse-title {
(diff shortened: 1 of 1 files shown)


saadeghi:
There is a problem about this. When clicking on the content (anything inside `collapse-content`) it also causes it to get closed. I don't think it should get closed.

Also, What if someone don't want it to get closed on second click?

kimdre:
Unfortunately I don't know a solution for that. I just wanted to propose a simple solution for that problem as a first step.

saadeghi:
I close this PR for now until I find a better solution. 

Thank you 💙

PR #1510, opened 2023-01-10
fix(docs): update and improve japanese translation

cohu-dev:
Hi

Some Japanese translations of the documentation were missing or inconsistent, so they have been corrected.

One of them is like below.

スクリーンショット 2023-01-10 13 45 23

And the following modifications are made.

image

This allowed us to add Japanese translations of all components.

Please check and merge when you have time.


saadeghi:
Thank you

PR #1529, opened 2023-01-16
unnecessary console logs removed

therealrinku:
removed redundant and unnecessary console logs


saadeghi:
Thank you but they make a new line. Without them, there will be no space between daisyUI logs and any other message on the terminal. To avoid confusion, I think it's necessary to have those new lines.

Issue #1530, opened 2023-01-16
Feature Request: Add Focus colors to all semantic colors


fini:
Add the following semantic colors:
- info-focus
- success-focus
- warning-focus
- error-focus

Primary, secondary and accent colors, have a 'focus' sibling color that is a darker tone.
It would be beneficial if that was also applied to info, success, warning and error.

That would enable designs to have a success / warning box with a fitting border colour etc.

fini:
I have added a small PR:
https://github.com/saadeghi/daisyui/pull/1531

saadeghi:
Thank you for the PR but I really don't want to add more color CSS variables especially for the state colors like `error`, `warning`, etc.  
I think even the existing CSS variable for `*-focus` colors are unnecessary because it's always the same color with darker tone. In almost every design system.  
Unfortunately we can't use CSS [`color-mix()`](https://caniuse.com/mdn-css_types_color_color-mix) yet so we have to add a new color value to get a darker tone of the same color. And just because they are new color values, I added CSS variables for them. I'm planning to remove all `*-focus` colors as soon as `color-mix()` is supported by major borwsers.

State colors (`error`, `warning`, `success`, `info`) should be used for validation or showing alert or info messages so I don't think they really need an alternative shade.

I apologize but I can't merge this PR. However it is possible to add any custom color to daisyUI and make it work with themes: https://github.com/saadeghi/daisyui/discussions/653#discussioncomment-2438428

fini:
I do understand if you don't feel it should be included, but to clarify my usecase:

Here are some examples of me using the those colors, where it would be nice if I could set the outline to outline-success-focus (darker shade)

![Screenshot 2023-01-25 at 16 43 10](https://user-images.githubusercontent.com/41289/214608225-06102be4-dcec-4623-ad89-d6b099d61c41.png)
![Screenshot 2023-01-25 at 16 42 50](https://user-images.githubusercontent.com/41289/214608229-805f7d95-0589-4b23-9376-7a8b10ac8ef2.png)

saadeghi:
There are many ways to customize. For eaxmple: https://play.tailwindcss.com/ckthfl0i9M?file=css

PR #1531, opened 2023-01-16
Add more semantic focus colors

fini:
Add the following semantic colors:

info-focus
success-focus
warning-focus
error-focus
Primary, secondary and accent colors, have a 'focus' sibling color that is a darker tone.
It would be beneficial if that was also applied to info, success, warning and error.

That would enable designs to have a success / warning box with a fitting border colour etc.

### DIFF
--- a/src/colors/colorNames.js
+++ b/src/colors/colorNames.js
@@ -21,14 +21,18 @@ module.exports = {
   "base-content": "--bc",
 
   info: "--in",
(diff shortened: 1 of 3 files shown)


saadeghi:
Thank you for the PR but unfortunately I can't merge it. Please read the reason below:  
https://github.com/saadeghi/daisyui/issues/1530#issuecomment-1403620960

PR #1532, opened 2023-01-16
update: added instructions for yarn install and copy to clipboard but…

therealrinku:
##DOCS
- Added instructions for yarn install in homepage(using tabs) and in install instructions page .
- Also added copy to clipboard button in homepage to copy yarn add daisyui/ npm i daisyUI scripts .

### DIFF
--- a/src/docs/src/components/homepage/Hero.svelte
+++ b/src/docs/src/components/homepage/Hero.svelte
@@ -1,6 +1,28 @@
 <script>
   import Ads from "@components/Ads.svelte"
   import { t } from "@src/lib/i18n"
+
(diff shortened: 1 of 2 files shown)


saadeghi:
Thank you for the PR but I really don't want to go in that road.  
Every year a new package manager gets released, there's `NPM`, `Yarn`, `PNPM`, `Bun` and there will be more. If a developer is using anything other than NPM, I'm sure they know how to install a package using the package manager of their choice.

therealrinku:
Hi @saadeghi . Actually that make sense. Thanks for the feedback .

Issue #1523, opened 2023-01-13
Feature request: Read only rating component.


Aglezabad:
Good evening:

I've checked the code and I've seen that rating component only uses _input_ tags because it's used for inputting ratings instead of showing the results. I need a read only variant for the latter.

My own try was using _input_ with disabled attribute and changing the cursor style but I think it's not the optimal way to do it, so I've made a read only variant of rating component using _span_ tag.

Requires some additions to the css files, which I've already made [in a fork](https://github.com/Japporters/daisyui/tree/readonly-rating).  I'm creating this issue for requesting the authorization to send a pull request.

Regards.

Here's an example of HTML with the proposal:
```html
<div class="rating">
  <span class="mask mask-star-2 bg-primary checked"></span>
  <span class="mask mask-star-2 bg-primary checked"></span>
  <span class="mask mask-star-2 bg-primary checked"></span>
  <span class="mask mask-star-2 bg-primary checked"></span>
  <span class="mask mask-star-2 bg-primary"></span>
</div>
```

Aglezabad:
For your convenience, I've created the pull request with changes required for the proposal.

Regards.

saadeghi:
Hey, Thank you for the PR but I think adding `disabled` attribute is actually the optimal way. If it is a data input, I think using a standard input element is better than using a `span` or `div` or anything else. and If it should be read only, just like any other input element, we can disable it.

If you don't want it to be input, you can just use the `mask` or any other icon and put them together. There's no need to use rating. Because rating is a data input.

Example: https://play.tailwindcss.com/a3cw853e6F

PR #1535, opened 2023-01-17
Feature request: Read only rating component.

Aglezabad:
Related with saadeghi/daisyui#1523 feature request, these are the changes made for the proposal.

In case of any change in code is required, let me know.

Regards.

### DIFF
--- a/src/components/styled/rating.css
+++ b/src/components/styled/rating.css
@@ -3,17 +3,21 @@
     @apply appearance-none;
     -webkit-appearance: none;
   }
+  
   :where(input) {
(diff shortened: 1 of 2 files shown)


saadeghi:
Thank you for the PR but I don't think it would be useful. adding a `disabled` attribute just does the job and if someone doesn't want to use inputs for some reason, they can just put some star icons next to each other. no need to use `rating`.  

My comment on the issue: https://github.com/saadeghi/daisyui/issues/1523#issuecomment-1386187352

PR #1538, opened 2023-01-17
Fix #1537: Add input-default class to reset input color

### LINKED ISSUE #1537: Feature Request: add `input-default` to reset input style back to default (unstyled background color)
This would be useful to e.g. turn an `input-success` back into a standard `input` based on a focus or hover selector.

Use case: `<input class="input input-success focus:input-default" ... />`

I've accomplished this with the following in my main tailwind css file:
```
@layer components {
  .input-default {
    border-color: hsl(var(--b1) / var(--tw-bg-opacity));
  }
  .input-default:focus {
    outline: 2px solid hsla(var(--bc) / 0.2);
    outline-offset: 2px;
  }
}
```

fnimick:
Also update the docs to show an example of resetting input style on focus.

---

@saadeghi I could not get the `focus:` variant to work correctly on the docs site in development mode, however, running it with `NODE_ENV=production` functions as expected.

I suspect this is because in development mode, the generated CSS files from daisyui `dist` are imported directly, without the focus variants being generated by the tailwindcss plugin. I attempted to resolve this by forcing generation in both `lib/responsiveRegex.js` and in the tailwind `safelist` in `docs/tailwind.config.cjs`, but neither worked. I couldn't figure out how the CSS is being imported in development mode in order to debug further.

### DIFF
--- a/src/components/styled/input.css
+++ b/src/components/styled/input.css
@@ -63,6 +63,13 @@
   &[disabled] {
     @apply cursor-not-allowed border-base-200 bg-base-200 text-opacity-20 placeholder-base-content placeholder-opacity-20;
   }
+  &-default {
+    @apply border-base-content;
+    &:focus {
+      outline: 2px solid hsla(var(--bc) / 0.2);
+      outline-offset: 2px;
+    }
+  }
   /* &::-webkit-calendar-picker-indicator {
     display: none;
   } */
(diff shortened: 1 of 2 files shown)


saadeghi:
Thank you for the PR but I don't think this will be useful.  
Reason: https://github.com/saadeghi/daisyui/issues/1537#issuecomment-1386169091

PR #1588, opened 2023-02-10
Fix small error in the description

castroalves:
The text in the description should be `primary` instead of `secondary`


saadeghi:
Thank you

PR #1605, opened 2023-02-15
tab-disabled

sfinktah:
Just a tab-disabled, as with other controls.  Sometimes you just need to disable a tab.


saadeghi:
thank you

PR #1626, opened 2023-02-21
fix: border-radius in now applied to all tabs, whether active or not (in tabs-boxed layout)

ffranchina:
Hello! I made a very simple and straightforward modification to fix a bug that I found in a project of mine.

**Description of the problem**: In the layout [Boxed Tabs](https://daisyui.com/components/tab/#boxed), the tabs had `border-radius` set only when _active_ (with the attribute `tab-active`).

The implications of such behavior are evident if, for instance, we set a different `background-color` for the other elements:
immagine
Such behavior is also quite annoying while working with transitions or other kind of animations.

**The solution**: The easy and clean solution is simply to apply the `border-radius` to all the `.tab` elements. Even if they are transparent, they will have the property set and they will behave accordingly even when colored or animated.

Thank you very much for such a great and clean project!! :heart:

### DIFF
--- a/src/components/styled/tab.css
+++ b/src/components/styled/tab.css
@@ -92,7 +92,10 @@
 }
 .tabs-boxed {
   @apply bg-base-200 p-1 rounded-btn;
+  .tab {
+    @apply rounded-btn;
+  }
   .tab-active {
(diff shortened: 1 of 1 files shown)


saadeghi:
thank you

Issue #1569, opened 2023-02-01
bug: breadcrumb should support an ordered list (SEO)


digitalcortex:
### What version of daisyUI are you using?

v2.49.0

### Describe your issue

I tried using breadcrumbs from [example page](https://daisyui.com/components/breadcrumbs/) but the source code is written with `<ul>`. Changing `<ul>` to `<ol>` didn't help probably because DaisyUI does not include style definitions for an ordered list inside the breadcrumb tag.

Using ordered list `<ol>` for breadcrumbs is semantically better and may be important for SEO. By using ordered list you are telling the search engine that the order of the links inside your breadcrumb is important. Ordered list for breadcrumbs creates a meaningful hierarchy for your internal navigation

Is there any way to update class definition to support `<ol>` tags as well?

amirsinaa:
Hi,
I've fixed this issue: #1648 
@saadeghi @digitalcortex

saadeghi:
Thank you @amirsinaa. This will be merged today

PR #1648, opened 2023-03-01
fix: #1569 breadcrumb should support an ordered list (SEO)

amirsinaa:
### What is fixed
breadcrumb should support an ordered list (SEO)
#1569 

Reproduction of the issue: https://play.tailwindcss.com/lWnURBWN1C


saadeghi:
Thank you.  
Will be merged today.

Issue #1666, opened 2023-03-06
bug: Button Group Responsive Issues


craigerskine:
### What version of daisyUI are you using?

v2.51.3

### Describe your issue

Button group responsive classes not working properly. Even the docs are showing it incorrectly.

![Screenshot 2023-03-06 at 9 51 43 AM](https://user-images.githubusercontent.com/56437/223162177-d08de8b9-5208-4382-94f4-59c12c24f7bc.png)

### What browsers are you seeing the problem on?

Chrome

### Reproduction URL (optional)

https://daisyui.com/components/button-group/#responsive---vertical-for-small-screen-horizontal-on-large-screen

nazaninreihani:
Please Assign this to me

saadeghi:
@nazaninreihani sure

nazaninreihani:
@saadeghi Thx

PR #1676, opened 2023-03-09
fix: #1666 Button group responsive classes not working properly.

nazaninreihani:


saadeghi:
I think this should be fixed in the CSS.

saadeghi:
Thank you

PR #1684, opened 2023-03-11
 change console.log to console.warn

ryoppippi:
I would like to change how to print the messages 
This is because, when we use console.log for output logs and use `prettier` for formatting files, it sometimes contaminates the files


ryoppippi:
Sorry I found that this is not tailwind issue, but my null-ls.nvim issue

PR #1685, opened 2023-03-11
`Респонсивный` -> `Отзывчивый`

scffs:
I think `Responsive` means that components and anything else adapt to different devices. In the Russian community, we call it `Отзывчивый дизайн` or `Адаптивный дизайн` (Responsive design or Adaptive design)


saadeghi:
Thank you

PR #1718, opened 2023-03-28
Fix `image-full` modifier to use the correct prefix in the card component

mixmav:
The code examples in the docs weren't using the reactive prefix for the `image-full` modifier.
This fixes the issue.


saadeghi:
Thank you

PR #1745, opened 2023-04-12
Refactor colorGen functions

gurgunday:
Hey @saadeghi! Since we're heavily using DaisyUI in production, especially the dynamic color generators, I've been wanting to optimize them for a while now. This is just an initial PR where I tried to make the code simpler, faster (mostly by reducing unnecessary method calls), and stricter in its checks.

There isn't any change in functionality here, and it outputs the same CSS in our codebase — tests pass too. You can check the commits one by one to see the modifications made to each function.

I know most people use DaisyUI in a CLI setting, but I think better performance is always good. Please let me know if you want any changes.

PR #1752, opened 2023-04-14
Refactor color functions

gurgunday:
👻

#1745


gurgunday:
This is not to v3... will fix.

saadeghi:
> This is not to v3... will fix.

Thank you. Because v3 is a new branch and when version 3 gets ready, it will be the default brach.

PR #1754, opened 2023-04-14
refactor color functions

gurgunday:
Oops, sorry for the mess, `git rebase` hates me 🤣

### DIFF
--- a/src/colors/functions.js
+++ b/src/colors/functions.js
@@ -163,7 +163,7 @@ module.exports = {
   },
 
   injectThemes: function (addBase, config, themes) {
-    const includedThemesObj = new Object()
+    const includedThemesObj = {}
 
     // add light themes
     if (!config("daisyui.themes")) {
@@ -251,8 +251,8 @@ module.exports = {
     })
 
     return {
-      includedThemesObj: includedThemesObj,
-      themeOrder: themeOrder,
+      includedThemesObj,
+      themeOrder,
     }
   },
 }
--- a/src/colors/functions.js
+++ b/src/colors/functions.js
@@ -191,7 +191,7 @@ module.exports = {
     }
 
     let themeOrder = []
-    if (!config("daisyui.themes") && config("daisyui.themes") !== undefined) {
+    if (config("daisyui.themes") === false) {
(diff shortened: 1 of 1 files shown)


saadeghi:
No problem. I should've mentioned before, that `v3` is a new (orphan) branch.  
Thank you 🙏

saadeghi:
There's a problem. It's only generating the default theme.

gurgunday:
Can you try again?

gurgunday:
I also wanted to ask you about that if check...

Should it only apply the light theme if `false` is given, or should it act that way for any falsy value (like null)? For now, I tried to mimic the previous behavior

saadeghi:
> Should it only apply the light theme if `false` is given, or should it act that way for any falsy value (like null)? For now, I tried to mimic the previous behavior

It should have the the exact same behavior as before: https://daisyui.com/docs/themes/#-2  
Because otherwise it would suddenly change every project that is using daisyUI and they have a specific config for their themes. I don't want to make breaking changes unless it's really necessary.

gurgunday:
Ok, now `injectThemes` behaves exactly the same way as before

polyzen:
> No problem. I should've mentioned before, that `v3` is a new (orphan) branch. Thank you pray

Why an orphan branch?

saadeghi:
To reduce the repo size. It was getting huge (almost 1.5GB) because of huge git history.

polyzen:
I'm not entirely against that, but this is a good read:
https://www.mail-archive.com/dri-devel@lists.sourceforge.net/msg39091.html

saadeghi:
> I'm not entirely against that, but this is a good read: https://www.mail-archive.com/dri-devel@lists.sourceforge.net/msg39091.html

Thank you. I'm not going to erase anything, it's just a new branch. The problem is daisyUI repo changed a lot from version 0.x to 1.x and then 2.x and the repo also includes the docs site which was Nuxt at a point, then it changed to SvelteKit 0.x and now SvelteKit 1.x A lot of git history is really unnecessary to keep in the default branch, specially for people who want to fork and run the repo. The full history will be available of course on the other branch.

PR #1776, opened 2023-04-27
fix: safari radio component transition

debugtheworldbot:
the transition of radio component doesn't work well at Safari.

https://user-images.githubusercontent.com/62830430/234827331-4ac1ad69-a0ad-49d9-84b2-12905740b53d.mov
video-to-gif-converter

removing this transition doesn't change any animation effect of radio.


saadeghi:
Thank you

PR #1779, opened 2023-04-29
feat: use for border radius -box instead of -xl so they can match with theme configuration

Wenish:
So the table uses the normal tailwind classes
```
 @apply rounded-tl-lg;
 @apply rounded-tr-lg;
 @apply rounded-bl-lg;
 @apply rounded-br-lg;
 ```

this  dosnt match with themes that dont have border radius

so i made that they use the daisyui -box modifier
```
@apply rounded-tl-box;
@apply rounded-tr-box;
@apply rounded-bl-box;
@apply rounded-br-box;
```
now the table border radius match with the theme config

### DIFF
--- a/src/components/styled/table.css
+++ b/src/components/styled/table.css
@@ -48,11 +48,11 @@
   :where(*:first-child) {
     :where(th, td) {
       &:first-child {
-        @apply rounded-tl-lg;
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you

PR #1783, opened 2023-04-30
feat: Add fill current for Button

oruman:
Use fill-current for SVG images in Buttons


saadeghi:
This is not a good idea because not all SVG icons need a fill color. some SVG icons are just lines and need stroke color. they won't look good if we set fill color for them automatically.

PR #1804, opened 2023-05-09
fix: btn-group right-to-left direction border radius

AmirHosseinKarimi:
This PR fixes the button group border-radius issue on the right-to-left pages.

Before:
image

After:
image

### DIFF
--- a/src/utilities/styled/button.css
+++ b/src/utilities/styled/button.css
@@ -1,58 +1,58 @@
 /* group */
 .btn-group {
   .btn:not(:first-child):not(:last-child) {
-    border-top-left-radius: 0;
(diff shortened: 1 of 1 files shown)


AmirHosseinKarimi:
According to using start/end direction for border-radius in this PR, Here is browser compatibility of these CSS properties:
[caniuse.com](https://caniuse.com/mdn-css_properties_border-start-end-radius) - [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/border-start-end-radius#browser_compatibility)

 I'm not sure it's compatible with DaisyUI browser compatibility conducts but looks like it is compatible with modern browsers.

PR #1923, opened 2023-06-05
fix: remove the invalid "for" property from the close button

uceumice:
It has probably been a typo, but "for" property is not applicable to the button element.


saadeghi:
Thank you

Issue #1520, opened 2023-01-12
Drawer: Sticky bottom nav doesn't behave properly


csechuan:
Sticky bottom nav doesn't behave properly when open drawer.

Reproduce:
Try scroll down a bit on main drawer content before open the drawer. You will notice the bottom nav will move upward as much as you scroll down in the main.  

[https://play.tailwindcss.com/S9zsrSPbBc](https://play.tailwindcss.com/S9zsrSPbBc)

cappe:
I noticed this too this morning. It's the following CSS style in TailwindCSS' @components layer:

```
.drawer-toggle:checked ~ .drawer-content,
.drawer.drawer-end > .drawer-toggle:checked ~ .drawer-content {
  transform:  translate(var(--tw-translate-x),var(--tw-translate-y))
              rotate(var(--tw-rotate))
              skewX(var(--tw-skew-x)) 
              skewY(var(--tw-skew-y))
              scaleX(var(--tw-scale-x))
              scaleY(var(--tw-scale-y));
}
```

I could not figure out the fix (with limited time) so ended up overriding it with `transform: none;` for now.

saadeghi:
Ok I found the reason: having a `fixed` element inside a `transform`ed element would make it fixed to the transformed element instead of the closest `relative` element. I didn't know that before today.  

Read more:
- https://stackoverflow.com/questions/15194313/transform3d-not-working-with-position-fixed-children/15256339#15256339
- http://meyerweb.com/eric/thoughts/2011/09/12/un-fixing-fixed-elements-with-css-transforms/

I really liked that little transition of the `drawer-content` when `drawer` opens but maybe I should remove that if it is causing issues like this 🥲

cappe:
@saadeghi TIL, thanks for providing more details! I really liked the transition too, it made the touch & feel pretty nice.

I'll try to find some time and motivation to dig into it.

PR #1988, opened 2023-06-12
fix: #1888 drawer now animates out

megagames-me:
Fixes bug #1888, `drawer-overlay`'s dimensions no longer collapse and `drawer-side` no longer disappears when the drawer is closed.

https://github.com/saadeghi/daisyui/assets/54014569/c285ba58-effe-44f5-9dbf-062f85c086ec

### DIFF
--- a/src/components/unstyled/drawer.css
+++ b/src/components/unstyled/drawer.css
@@ -2,7 +2,7 @@
   @apply grid relative;
   grid-auto-columns: max-content auto;
   &-content {
-    @apply row-start-1 col-start-2 translate-x-0 transition-all;
+    @apply row-start-1 col-start-2;
   }
   &-side {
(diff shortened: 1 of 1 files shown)


saadeghi:
> I added a slight offset to the `drawer-content` like v2 had. I can remove this if needed though.

Yes please remove that translate because that transform causes [other issues](https://github.com/saadeghi/daisyui/issues/1520#issuecomment-1403551439)

If anyone wants that small effect, they can have it by adding a few class names like 
```
transition-transform translate-x-0 [.drawer-toggle:checked~&]:translate-x-2
```

megagames-me:
> Yes please remove that translate because that causes [other issues](https://github.com/saadeghi/daisyui/issues/1520#issuecomment-1403551439)

Ok, removed. Can you make sure the transition timings are correct though? They look a little off to me when it closes. I can change them if needed.

saadeghi:
Thank you

PR #1993, opened 2023-06-13
fix #1992: add/fix anims and improve docs for collapse

### LINKED ISSUE #1992: bug: issues with collapse & docs

v3.1.0

### Describe your issue

There seem to be multiple problems with the collapse component.

## Animation on collapse with custom background

When a custom background is applied to a collapse component, the border radius becomes messed up on the bottom area when it is transitioning to the collapsed state. This is also shown in the documentation but I have provided a simplified version of it.

## Text doesn't animate out properly

When closing any collapse components, the text disappears straight away instead of animating out. Collapse always has seemed to be rough and broken but I think fixing this would make it much smoother.

## `collapse-title` isn't vertically aligned sometimes

If the font size of `collapse-title` is too small, the text is too high up in the `collapse-title` container. This could be fixed with flexbox or vertical-align. Flexbox breaks in the animation, though, and vertical-align doesn't work properly with text that is big enough to fit. I'll see if I can fix it, though.

## Documentation inconsistency

For [this part of the documentation](https://daisyui.com/components/collapse/#custom-colors-for-collapse-that-works-with-checkbox), it says to use "Tailwind CSS `peer` and `peer-checked` utilities to apply style when sibling checkbox is checked" but the example doesn't use `peer-checked`. This should be an easy fix I think.

<hr />

I'd like to fix these issues myself if possible but I'd just like to document this if anyone has any ideas.

Chrome

### Reproduction URL (optional)

https://play.tailwindcss.com/hh0gQWQsir

megagames-me:
Added animations and fixed the border radius bug on the collapse component when it closes. I also changed [this part of the documentation for collapse](https://daisyui.com/components/collapse/#custom-colors-for-collapse-that-works-with-checkbox) to use `peer-checked` like it suggests in the info box above the code box.

Example of new behavior:

https://github.com/saadeghi/daisyui/assets/54014569/7a1e96a4-920f-40d0-873d-746a7501c729

### DIFF
--- a/src/components/unstyled/collapse.css
+++ b/src/components/unstyled/collapse.css
@@ -37,7 +37,7 @@
 .collapse:focus:not(.collapse-close) .collapse-content,
 .collapse:not(.collapse-close) input[type="checkbox"]:checked ~ .collapse-content,
 .collapse:not(.collapse-close) input[type="radio"]:checked ~ .collapse-content {
-  @apply visible min-h-fit h-auto translate-y-0;
+  @apply visible min-h-fit translate-y-0;
   & > * {
     @apply translate-y-0;
     transition: transform 0s ease-in-out;
--- a/src/components/unstyled/collapse.css
+++ b/src/components/unstyled/collapse.css
@@ -17,8 +17,11 @@
   @apply appearance-none opacity-0;
 }
 .collapse-content {
-  @apply col-start-1 row-start-2 invisible min-h-0;
-  transition: visibility 0.2s;
+  @apply col-start-1 row-start-2 min-h-0;
+  & > * {
+    @apply translate-y-4;
+    transition: transform 0.2s ease-in-out;
+  }
 }
 .collapse[open],
 .collapse-open,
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks for the PR but I don't see how changes in `collapse.css` file are improving anything...?  
- `.collapse-content > *` is not a good idea at all. We shouldn't apply transform or any other styles to **all** elements inside `collapse-content` because someone may have other styles for their elements. We should only apply styles to the class names we provide.  
- Why add `translate-y-0` to `.collapse-content`?
- Why remove `invisible` class from `collapse-content`?
- Why remove `transition` from `collapse-content`?

None of the changes in `collapse.css` make sense to me but I appreciate the changes in `collapse/+page.svelte.md`. If you reset the CSS file, I can merge the changes in `collapse/+page.svelte.md`

Thanks

megagames-me:
Hi, thanks for the feedback. It could be that the main issue was specific to my system because [this example](https://daisyui.com/components/collapse/#custom-colors-for-collapse-that-works-with-focus) is quite broken for me. Removing the invisible class fixed most of the issues with it. If this doesn't work, I'm not sure how to fix it because the behavior with animating `grid-template` seems kind of broken as the top row stretches out a bit when animating. When animating out (this animation going out is created by removing the invisible class, I'm not sure whether it would break anything though) without the translate, there was some sort of flicker because of the weird `grid-template` animation behavior.

I'll remove the changes to collapse.css but I think the issue still remains so I'll keep the issue open. Hopefully someone else finds a better solution to this.

influencep (approved):

influencep (approved):

saadeghi:
https://github.com/saadeghi/daisyui/issues/1992#issuecomment-1600716477

megagames-me:
@saadeghi I've removed the changes

saadeghi:
Thank you

Issue #1995, opened 2023-06-13
docs: invalid "for" property from the close button


fabianmossberg:
### On which page do you see this issue?

https://daisyui.com/components/modal/

### Describe the issue

The examples of the modal has an invalid "for" property for the close button

### What browsers are you seeing the problem on?

All browsers

PR #1996, opened 2023-06-13
fix: #1995 invalid "for" property in examples

fabianmossberg:
In the PR #1923, this issue was fixed in the component itself.

This PR also removes it from the example code.


saadeghi:
Thank you

PR #2020, opened 2023-06-19
fix docs for customizing themes

uluumbch:
RIght now the docs for "How to customize an existing theme?" and  "How to add custom styles for a specific theme?" says we have to require `...require("daisyui/src/theming/themes")["[data-theme=light]"]` but this line give me an error, so I fix this error by changing to `...require("daisyui/src/colors/themes")["[data-theme=light]"]` 

Screenshot for the error
image
image

### DIFF
--- a/src/docs/src/routes/docs/themes/+page.svelte.md
+++ b/src/docs/src/routes/docs/themes/+page.svelte.md
@@ -230,7 +230,7 @@ module.exports = {
     themes: [
       {
         light: {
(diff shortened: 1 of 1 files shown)


saadeghi:
I think you're using daisyUI version 2.x  
The directory name changed from `colors` to `theming` in version 3.

- https://unpkg.com/browse/daisyui@3.1.1/src/  
- https://unpkg.com/browse/daisyui@2.52.0/src/

Use the v2 docs if you're not going to upgrade to v3 yet:  
https://v2.daisyui.com/

PR #2072, opened 2023-06-29
docs(indicator): Mismatch preview and html

yoshi6jp:
https://daisyui.com/components/indicator/#multiple-indicators


saadeghi:
Thank you

Issue #2087, opened 2023-07-02
bug: In rtl mode, the checkbox icons are reversed


parsgit:
### What version of daisyUI are you using?

3.1.7

### Describe your issue

In RTL mode, the check box icon is not shown correctly. In this case, only the position of the elements should be changed

https://play.tailwindcss.com/yTsrcbPaFJ

### What browsers are you seeing the problem on?

Chrome, Firefox

### Reproduction URL

https://play.tailwindcss.com/yTsrcbPaFJ

saadeghi:
From docs: https://daisyui.com/docs/config/#rtl

>You need to add `dir='rtl'` to your body tag. If you're using daisyUI with RTL option, I suggest using [tailwindcss-flip](https://github.com/cvrajeesh/tailwindcss-flip) plugin to flip all your Tailwind utilities automatically.

PR #2088, opened 2023-07-02
Fix tick icon in RTL mode

parsgit:
Fix this (#2087)


saadeghi:
https://github.com/saadeghi/daisyui/issues/2087#issuecomment-1618184942

PR #2120, opened 2023-07-10
feat: add theme extraction helper

Goldziher:
This PR adds a utility function to extract the theme from the DOM, as discussed here: https://github.com/saadeghi/daisyui/discussions/415#discussioncomment-6406762

I'm afraid I didn't add documentation for this, since I have no idea how the documentation works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "!dist/*.rtl.css",
     "src/index.js",
     "src/theming/*.js",
+    "src/helpers/*.*",
     "src/index.d.ts"
   ],
   "engines": {
--- a/src/helpers/index.d.ts
+++ b/src/helpers/index.d.ts
@@ -1,0 +1,29 @@
+export interface ThemeColors {
+    primary: string;
+    primaryFocus: string;
+    primaryContent: string;
+    secondary: string;
+    secondaryFocus: string;
+    secondaryContent: string;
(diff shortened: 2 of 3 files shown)


saadeghi:
Oh, I thought you just wanted to add types for Typescript. Sorry.

I'm not sure about the JS file that returns the color values of the current theme. It might be helpful for some people but to use something like this, I think daisyUI should be installed as a normal dependency (instead of dev dependency)  

And I'm not sure how percentage of daisyUI users may need something like this if we're adding this file for everyone...

saadeghi:
My bad, I saw your comment from the original discussion on my phone and I thought it's about adding types for the color variables (so editor's intellisense shows the types for each CSS variable) 🫡

Thank you for your time and effort. For now I think daisyUI doesn't need to be a normal dependency. It just works as a plugin for Tailwind CSS at build time. So including helper functions like this inside the package is not a little bit too extra 😅

Thank you and sorry about my misunderstanding

cotontigeh:
Thanks for this PR @Goldziher, i implement your scripts on my project and it's work like a charm ! 

It's very useful in my case with [React Select](https://react-select.com/styles#overriding-the-theme) Theming where i want to retrieve the colors from current theme to override the select theme

Thanks again ! 😍

PR #2129, opened 2023-07-13
docs: make some paragraph and alert translatable

5idereal:
Also updated the Traditional Chinese translation.


saadeghi:
Thank you

PR #2150, opened 2023-07-19
remove taiwan flag

92hackers:
Hi, @saadeghi ,Thanks for your hard work to bring this awesome css lib to us developers.

This PR is just a fix to daisyui docs website, which make a mistake at TaiWan flag.

TaiWan is not a country, is essentially a province of China. currently it's a district because of some history reasons.

So, TaiWan flag should not be placed along with other REAL country flags, It's a big misleading to website visitors.

----------------

Besides, 繁体中文，not only spoken by taiwan people, but also HongKong people, Macao people and many overseas-chinese people.
so taiwan flag can not represents 繁体中文 at all. 

----------------
My fix is to hidden taiwan flag, and preserve 繁体中文 name. below are screenshots showing brefore and after views:

Before:

Screen Shot 2023-07-19 at 5 55 56 PM

After:

Screen Shot 2023-07-19 at 6 47 25 PM

### DIFF
--- a/src/docs/src/components/LangChange.svelte
+++ b/src/docs/src/components/LangChange.svelte
@@ -34,7 +34,7 @@
         {#if $t("__name", {}, langItem, false) !== "__name"}
           <li>
             <button class:active={$currentLang == langItem} on:click={() => setLang(langItem)}>
-              {#if $t("__flag", {}, langItem, false) !== "__flag"}
+              {#if $t("__flag", {}, langItem, false) !== "__flag" && $t("__name_en", {}, langItem, false) !== "taiwan"}
                 <img
                   class="drop-shadow"
                   loading="lazy"
--- a/src/docs/src/translation/zh_tw.json
+++ b/src/docs/src/translation/zh_tw.json
@@ -1,5 +1,6 @@
 {
   "__name": "繁體中文",
+  "__name_en": "taiwan",
   "__flag": "🇹🇼",
   "__status": "",
   "Tailwind CSS Components": "Tailwind CSS 元件",


saadeghi:
> Hi, @saadeghi ,Thanks for your hard work to bring this awesome css lib to us developers.

Thank you 🙏

>TaiWan is not a country

I have zero knowledge about the Chinese language. I know there are political disagreements but I would appreciate if you don't make a CSS library involved in politics.  

The translation and the flag was added by people who are living there and I think you would agree that removing the flag would also lead people who are disagree with you, to get angry. 

I really don't want to get into politics so I appreciate if you don't insist on this 🙏



> Besides, 繁体中文，not only spoken by taiwan people, but also HongKong people, Macao people and many overseas-chinese people.
> so taiwan flag can not represents 繁体中文 at all.

This is also true about other flags. A lot of countries speak English, Portuguese, Spanish, etc. The flags are just the commonly accepted icon for that. I understand that a language is not limited to one region or country.


Thank you for understanding

NikarashiHatsu:
I think it is unnecessary to remove the flag, besides the political reasons, it hurts the visual integrity.

ikudev:
@92hackers Taiwan is definitely a country. It uses a different passport from the China mainland. It's ridiculous to say Taiwan is a province of China. Are Chinese people able to go to Taiwan without visa? Fool yourself but don't fool everybody else, OK?

saadeghi:
I just want to add some information here.

Screenshot_2023-07-21-00-35-38-331_com android chrome

These are the language files and the file names are based on the ISO standards.  
Most of the languages were added by open source contributors.  
The file name for the mentioned language is `zh_tw` which is specifically for the Chinese language that people use in Taiwan (whether you believe it is a real country or not). If this translation is for Chinese speakers from Taiwan area, I don't see any issues with the flag because flags here are not representing countries.  
Whether anyone believes Taiwan is a country or not (I don't care personally, sorry) no one can deny that it owns a flag.

PR #2156, opened 2023-07-21
docs: ✏️ refactor language menu

ThaddeusJiang:
Taiwan politics aside, having country flags associated with languages is always a bad idea. 

For Example, English and Spanish not associated with US and Mexico flags.


chouhanyang:
Well, like people said, without those flags, it turns the menu to an ugly piece of UI.
Plus, your PR is misleading, as your change simply remove all flags. Not splitting two things as advertised. 

I'd say, keep them as it is, and for those who want to change/remove, it would be a lot easier to change in their private clone than adding them back for all languages. :)

ThaddeusJiang:
@chouhanyang really?

Why Google and Apple productions use `the ugly` UI? 

Chrome Settings > languages
CleanShot 2023-07-21 at 15 15 14@2x


macOS Settings > languages
CleanShot 2023-07-21 at 15 15 03@2x

92hackers:
@ThaddeusJiang , my opinion, language is just language,  country flag means specific that country. 

There is unnecessary to bind language and country together.

In the context here, what we want is just language, and please do not bring political issues in.

ThaddeusJiang:
@92hackers Don't do [this](https://github.com/saadeghi/daisyui/pull/2150) any more, please.

saadeghi:
I'm not against hiding all flags.  
I understand the criticism about connecting flags to countries in general

However this PR should be on hold for now because if it get's merged today, people see this as a response or taking sides about the other PR and its political disagreement.

christorange:
> In the context here, what we want is just language, and please do not bring political issues in.

It's so funny this comes from someone said "TaiWan is not a country, is essentially a province of China. " 🤡 It can't be more obvious who's bringing in political flights here.

chouhanyang:
Not sure why you posted the screenshot to disapprove your point. Bunch of languages are shown with country names. Languages, no matter you like it or not, develop in different countries, as the Chinese use is Hong Kong needs to listed separately to distinguish itself a different language in terms of UI. 
image

Regardless, your change does not reflect the title of the PR, hence it is very misleading and dishonest. This code change simply removes all flags without "refactoring" or "splitting" them to separate places. A better option would be to provide options for people to show/hide flags. A library without giving people freedom to choose is subpar.

lcmgh:
Not everyone lives in a place in that country borders correlate so strongly with linguistic borders. https://cognitiveitching.wordpress.com/2014/02/11/linguistic-itch-the-real-borders-of-the-world-language/

saadeghi:
Thanks for the PR but I had to make some additional changes.  
Language dropdown now shows the standard ISO language codes instead of flags.

PR #2161, opened 2023-07-21
docs: update id.json and accordion examples to be able to support translation

mahdihrs:
Update Indonesia translation due to components update


NikarashiHatsu (commented):
LGTM

PR #2168, opened 2023-07-23
added :not(.join)

sfinktah:
Allows for the "pulling out" of a `join-item` from the border-radius first/last rounding system, by using nested `join`s.

_e.g.:_

`before`
image

`after`
image

```html
<div class="join">
    <button class="btn pointer-events-none">Types</button>
    <div class="join">
        <input class="join-item btn btn-outline" type="checkbox" aria-label="Shares">
        <input class="join-item btn btn-outline" type="checkbox" aria-label="Photos">
        <input class="join-item btn btn-outline" type="checkbox" aria-label="Videos">
    </div>
</div>
```

### DIFF
--- a/src/components/unstyled/join.css
+++ b/src/components/unstyled/join.css
@@ -7,7 +7,7 @@
     border-top-left-radius: 0;
   }
   & .join-item:not(:first-child):not(:last-child),
-  & *:not(:first-child):not(:last-child) .join-item {
+  & *:not(:first-child):not(:last-child):not(.join) .join-item {
     border-top-right-radius: 0;
     border-bottom-right-radius: 0;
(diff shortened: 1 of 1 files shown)


saadeghi:
If you don't need the first item `Types` to connect to the others, why did you use a `join` class for the parent?  
It could be: 
```html
<div>
  <button class="btn pointer-events-none">Types</button>
  <div class="join">
    <input class="join-item btn btn-outline" type="checkbox" aria-label="Shares">
    <input class="join-item btn btn-outline" type="checkbox" aria-label="Photos">
    <input class="join-item btn btn-outline" type="checkbox" aria-label="Videos">
  </div>
</div>
```

PR #2181, opened 2023-07-27
Adding default fade for toast

writeDavid:
Adding a default fade for a toast

### DIFF
--- a/src/components/styled/toast.css
+++ b/src/components/styled/toast.css
@@ -1,7 +1,7 @@
 .toast {
   @apply gap-2 p-4;
   & > * {
-    animation: toast-pop 0.25s ease-out;
+    animation: toast-pop 0.25s ease-out, toast-fade 0.25s 2.80s ease-out;
   }
 }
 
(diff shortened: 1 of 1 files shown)


writeDavid:
You've stated that you don't want to impose logic with the library.  Maybe new var(s) for the pop and fade?

saadeghi:
What does this animation do?  
Is it for when you hide/remove an item? Because that won't work. Hiding or removing is instant and it happens before the CSS animation.

writeDavid:
When a toast is triggered, your pop animation runs.  It then waits 2.8 seconds, and then automatically fades it with the next animation (the added code).

saadeghi:
First problem, It doesn't seem to work as you explained. It shows up with the `toast-pop` animation, then after `2.8s` it hides with the new `toast-fade` animation but immediately (as soon as animation ends - 0.25s) it becomes visible again.  

Second problem, what if someone doesn't want the toast to auto hide?  
Maybe they want it to be closed on click or using a JS timer.  
I think if we're adding something like this, it should be with new class like `auto-hide`

writeDavid:
> Second problem, what if someone doesn't want the toast to auto hide? Maybe they want it to be closed on click or using a JS timer. I think if we're adding something like this, it should be with new class like `auto-hide`

The code was pulled from a working implementation, that handles toasts via a svelte store, so apologies I didn't forsee the issue.  I'll close this, because you're right, this type of request is better left to project implementors.

Thank you for the feedback!

PR #2189, opened 2023-07-29
daisyui/src/theming -> daisyui/src/colors

gustawdaniel:
Latest daisyui use colors directory instead of theming


saadeghi:
> Latest daisyui use colors directory instead of theming

That's not true:  
https://unpkg.com/browse/daisyui@3.5.0/src/

I think you're using an old version.

PR #2255, opened 2023-08-16
fix: clarify description of customizing component styles

anewman15:
I felt the need for a clarification because daisyUI provides only assemblies of Tailwind classes for an intended component. The intended component, for example a button, is not an actual React component with APIs representing any props or returned object - as one would expect from an Ant Design [`<Button />`](https://ant.design/components/button) or a PrimeUI [`<Button />`](https://primereact.org/button/).

daisyUI is not providing us React components like a `<Button />`, but just assembled style classes **intended** for a JSX element we would like to use them on. It is important to make that distinction. I wanted to clarify that.

### DIFF
--- a/src/docs/src/routes/docs/customize/+page.svelte.md
+++ b/src/docs/src/routes/docs/customize/+page.svelte.md
@@ -1,5 +1,5 @@
 ---
-title: Customize daisyUI components
+title: Customize daisyUI component styles
 desc: How to customize daisyUI?
 published: true
 ---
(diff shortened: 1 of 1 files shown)


saadeghi:
>The intended component, for example a button, is not an actual React component

Where is it claimed that daisyUI is providing React components?

The word "component" doesn't mean a React component. We had CSS component in Bootstrap and other CSS libraries way before React existed.

PR #2259, opened 2023-08-17
feat(collapse): inherit focus state from children via `:focus-within`

ctjlewis:
With simple `:focus` selector, focusing on content (e.g. an input) will cause the box to collapse:

```html
<div class="collapse">
  <div class="collapse-title">
    <h1>Collapse</h1>
  </div>

  <div class="collapse-content">
    <!-- trying to type in the text input will close the container -->
    <input type="text">
  </div>
</div>
```

Instead, we can use `:focus-within` to inherit focus state from children as well. This prevents focus on children nodes from triggering the collapse. [It's fully supported across all browsers.](https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-within)

### DIFF
--- a/src/components/styled/collapse.css
+++ b/src/components/styled/collapse.css
@@ -48,7 +48,7 @@ details.collapse summary {
 .collapse:not(.collapse-open):not(.collapse-close) .collapse-title {
   @apply cursor-pointer;
 }
-.collapse:focus:not(.collapse-open):not(.collapse-close):not(.collapse[open]) .collapse-title {
+.collapse:focus-within:not(.collapse-open):not(.collapse-close):not(.collapse[open]) .collapse-title {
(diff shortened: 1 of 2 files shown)


ctjlewis:
@saadeghi Thank you for this library, it is the best component library available IMO. Please let me know what you think about this PR - I will have to hack it into my own project in the meantime, and respectfully recommend testing and merging the change.

saadeghi:
Thank you.  
This PR breaks [`Collapse with checkbox`](http://localhost:3000/components/collapse/#collapse-with-checkbox)  
With `:focus-within` the collapse with checkbox, opens with first click but wont close with second click. It needs the second click to uncheck the checkbox and then it needs another click outside to lose the `:focus-within` state.

PR #2286, opened 2023-08-22
Added themes shim

nandi95:
This themes.d.ts needs to end up next to the themes.js
Not sure which script to run to build the library and see what gets packed

The declaration merging as per https://github.com/saadeghi/daisyui/pull/2264 I still could not get it to work.


saadeghi:
> Not sure which script to run to build the library and see what gets packed

```
npm pack
```
New files won't be in the package unless you add them to [the `files` array here on `package.json`](https://github.com/saadeghi/daisyui/blob/b5e94ac42492c008f18fb16b3cfbb25ba7cbf50f/package.json#L36)

saadeghi:
> The declaration merging as per #2264 I still could not get it to work.

I don't use typescript but if there's something I can help with, let me know 🙏

nandi95:
This should be good to go now.

> I don't use typescript but if there's something I can help with, let me know 🙏


Just me or someone has to keep trying to get it done.
(even-tough you're not using ts, you could still benefit from this by using `/** @type {import('daisyui').Config} */` 😉

PR #2294, opened 2023-08-25
Footer: Change <div> tags to more "semantic" tags

danawoodman:
Changing `<div>` to use `<nav>`, `<aside>` and `<fieldset>` for containers and  `<span>` to `<h1>` for section headings.


polyzen:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements#avoid_using_multiple_h1_elements_on_one_page

saadeghi:
Thank you.  
The only problem is using `<h1>` tags.  Ideally those headings should be `<h*>` but it really depends on the content. For example, they could be `<h3>` on a website that uses `<h1>` as the main heading and `<h2>` for section titles. But they could be `<h6>` on a website that is already using `<h4>` and `<h5>` for the sections. I think if it's going to be semantic tag, it depends on the page hierarchy and we can't decide which `<h*>` tag people should use.

danawoodman:
We could go with a `<header>` tag which would perhaps be more semantically accurate. That said, I'm pretty sure using `<span>` is worse, esp considering it is using an inline tag as a block level element.

saadeghi:
> We could go with a `<header>` tag

Yes that would be good. Can you update the PR please?

danawoodman:
@saadeghi done!

saadeghi:
Thank you

bempensato:
Hi, this commit introduced a wrong approach since as specified by w3c (https://www.w3schools.com/tags/tag_header.asp) and mdn (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#technical_summary) <header> tags cannot be placed inside a <footer> element, should consider using divs?

danawoodman:
You're right @bempensato, I missed that particular rule, new PR up changing to `h6` over in #2791

bempensato:
@danawoodman @saadeghi  thank you for your superfast resolution!

PR #2339, opened 2023-09-12
zebra-alt (odd rows)

sfinktah:
Sorry, but even rows looks really wierd sometimes.

image

Behold... `<table class="table table-zebra-alt">`

image

### DIFF
--- a/src/components/styled/table.css
+++ b/src/components/styled/table.css
@@ -8,9 +8,9 @@
   &-zebra tbody tr:nth-child(even) {
     @apply bg-base-200;
   }
-  tr.hover,
-  tr.hover:nth-child(even) {
-    @apply [@media(hover:hover)]:hover:bg-base-200;
+
+  &-zebra-alt tbody tr:nth-child(odd) {
+    @apply bg-base-200;
   }
 
   &-zebra {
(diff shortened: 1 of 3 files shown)


saadeghi:
I don't think it's weird.  
Coloring odd instead of even doesn't fix any issue or provide any advantage as far as I know.
However you can always customize and extend classnames for your own project.

And, please create an issue or discussion before opening a PR to avoid unnecessary effort.  
((Contribution guideline)[https://github.com/saadeghi/daisyui/blob/master/.github/CONTRIBUTING.md#pull-requests])

sfinktah:
I'll admit it doesn't look wierd now that I look at it (a day later).  Though the is something about the tables that is a little sub-par compared to your other components which are all exquisitely designed.  I've never felt it necessary to make a visual change before.

I'll play around with it, perhaps darkening the headers and lightening the stripes.

How does one achieve such specific customizations while staying within the daisyui framework?  I note the shading (for example) does not lend itself to easy adjustment via `--var`, being set to `@apply bg-base-200;`

saadeghi:
> How does one achieve such specific customizations while staying within the daisyui framework? I note the shading (for example) does not lend itself to easy adjustment via `--var`, being set to `@apply bg-base-200;`

What customization are you looking for exactly? Changing the background of a row?

sfinktah:
> > How does one achieve such specific customizations while staying within the daisyui framework? I note the shading (for example) does not lend itself to easy adjustment via `--var`, being set to `@apply bg-base-200;`
> 
> What customization are you looking for exactly? Changing the background of a row?

Lets start with that... if it were only tailwindcss, I could do

```css
@layer utilities {
  .striped div:nth-child(even) {
    @apply bg-gray-100
  }
}
```

saadeghi:
I mean, what customization are you looking for about daisyUI tables?  
Because daisyUI is using something similar for zebra table
```
.table-zebra tbody tr:nth-child(2n)
```

sfinktah:
@saadeghi Sure... so if I wanted to use the alternate rows, how could I (as you said originally):

> However you can always customize and extend classnames for your own project.

An easier example would be if I needed buttons that weren't rounded, (which I actually do need).  Rather than write "btn rounded-0" everywhere, how could I apply that automatically?

PR #2352, opened 2023-09-16
add stat-sm

z3nful:
Added per my use case

### DIFF
--- a/src/components/styled/stat.css
+++ b/src/components/styled/stat.css
@@ -4,6 +4,12 @@
 :where(.stats) {
   @apply divide-x divide-y-0 overflow-x-auto;
 }
+.stats-sm {
+  @apply bg-base-100 text-base-content rounded-box flex;
+}
+:where(.stats-sm) {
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you for the PR,
However I think there are things that must be reconsidered.

1. daisyUI components that have size variants come with 4 sizes: `xs`, `sm`, `md` (default), `lg` so if stat is going to have different size variants I think it should be available on other sizes as well, not just `sm`
2. `stats` is the parent and `stat` is the item. I think if the size should be applied to the parent (`stats-sm`, not `stat-sm`) because there's no point of having `stat` items with different sizes, and if `stat` has size, it forces us to repeat the size class name for each item which is just extra.
3. parts of `stat` like `stat-title`, `stat-value`, `stat-desc` don't need to have `sm-` prefix at all. Ideally if it's `stats stats-sm` everything should be sized accordingly.

z3nful:
I understand, and I definitely jumped the gun on this a bit. I'll research some of the already built out components that use size variants a bit more and and see if I can get this more in line with current specs.

PR #2354, opened 2023-09-18
Add themeSelector for shadow root use cases

akx:
If variables are attached to `:root`, they don't seem to correctly transcend a shadow root's boundaries. I hacked my local daisyUI installation to attach the variables to `*` (like Tailwind itself does), and that makes the variables accessible for all elements.


rikgirbes:
Any update on this? Can this be merged?

saadeghi:
It will be merged in the next major version, alongside some improvements.  
Estimated release time is 2~3 weeks from now.

Issue #1982, opened 2023-06-12
bug: 'cat' is not recognized as an internal or external command on Windows when running the `build` command


kirklin:
### What version of daisyUI are you using?

v3.1.0

### Describe your issue

# 'cat' is not recognized as an internal or external command on Windows

## Description
When running the `build` command on Windows, the following error is thrown:

```
'cat' is not recognized as an internal or external command, operable program or batch file.
```

## Steps to reproduce
1. Clone the repository on a Windows machine
2. Run the `build` command in the terminal

## Expected behavior
The `build` command should compile all files successfully.

## Actual behavior
The `build` command throws an error due to the `cat` command not being recognized on Windows.

## Possible solutions
One solution is to use `type` or `copy` instead of `cat`, since `cat` is a UNIX-style command and is not available on Windows. Another solution is to use a build tool, such as Webpack or Gulp, to handle the compilation process and ensure cross-platform compatibility.

## Environment
- Operating system: Windows 11

Please let me know if there is any additional information I can provide. Thank you for your attention to this issue.

### What browsers are you seeing the problem on?

All browsers

saadeghi:
Can you use `Windows PowerShell`? I think `cat` command works there.  
(I don't really want to add a new dependency or a script just for OS specific commands like this one)

flomero:
unfortunately Powershell doesn't support `cat`
for everyone working with Windows: you can install and use `Git Bash` from [Git for Windows](https://gitforwindows.org/) and everything should run smoothly 🙌🏼

kirklin:
True, but I'd still like to have a set of scripts that can be used across all platforms

PR #2373, opened 2023-09-26
update CONTRIBUTING.md

flomero:
- updated the folder from `dist` to `src` according to https://github.com/saadeghi/daisyui/discussions/2370#discussioncomment-7107778
- added note for Windows Users having problems with `cat` https://github.com/saadeghi/daisyui/issues/1982


saadeghi:
Thank you

PR #2438, opened 2023-10-23
docs

devnio:
Default tailwind setup seems to be:

export default {
  content: [],
  theme: {
    extend: {},
  },
  plugins: [],
}


saadeghi:
https://tailwindcss.com/docs/installation  
Screenshot 2023-10-24 at 00 18 11

devnio:
Oh I was on the [installation page of sveltekit](https://tailwindcss.com/docs/guides/sveltekit). I guess in some cases it's different (for example also when using vite).
Maybe adding a small note below the code snippet saying that it might be different might help.

Issue #2408, opened 2023-10-13
bug: join-item impact the radius of its first and last children


meritozh:
### 1. Checklist

- [X] I updated daisyUI (`npm i -D daisyui@latest`) but I still have this issue
- [X] I looked up into [FAQ](https://github.com/saadeghi/daisyui/discussions/1949) but I couldn't find a solution
- [X] I searched in [existing issues](https://github.com/saadeghi/daisyui/issues?q=is%3Aissue) and [discussion forum](https://github.com/saadeghi/daisyui/discussions?discussions_q=) but this bug has not addressed before
- [X] This bug is dasiyUI issue and it's **NOT** specific to a JS framework or a build tool."

### 2. What version of daisyUI are you using?

v3.9.2

### 3. Which browsers are you seeing the problem on?

Chrome

### 4. Reproduction URL

https://play.tailwindcss.com/zemYLtPZQl

### 5. Describe your issue

Hey there, I find https://github.com/saadeghi/daisyui/issues/1957 closed, but in the latest version of DaisyUI, the join-item class still override the radius of **the first and last children**.

I think it should only impact it self border, not any children, which means below style isn't necessary **or make it optionally**?
https://github.com/saadeghi/daisyui/blob/48fc58f98ee96d62052f3393ad6a338a027eb854/src/components/unstyled/join.css#L47-L49

saadeghi:
There's a challenge here.

The reason we're targeting all items inside `join` using `.join *` is that `join-item` can be nested in divs and still should have radius, because of the inherited radius.  
This should be default behavior (not optional) because people use `join-item` anywhere inside `join` and they expect it to work.  

The issue can be fixed using something like `:where(.join *:has(.join-item))` instead of `:where(.join *)`  
But`:has` still is not supported in FireFox by default.

So if you have another solution for it, let me know.

Or you can just make a PR replacing
```css
:where(.join *) { 
 @apply rounded-[inherit]; 
}
```
with
```css
/* for firefox */
@supports not selector(:has(*)){
  :where(.join *){
    @apply rounded-[inherit];
  }
}

/* for others */
@supports selector(:has(*)){
  :where(.join *:has(.join-item)){
    @apply rounded-[inherit];
  }
}
```

PR #2441, opened 2023-10-24
no round nested join items

meritozh:
as described here: https://github.com/saadeghi/daisyui/issues/2408

tested in Chrome and Safari in macOS


saadeghi:
Thank you

PR #2453, opened 2023-10-31
zh_hans and zh_hant rename

5idereal:
The zh_hans and zh_hant translation files were interchanged. This pull request renamed the two files.


saadeghi:
Thank you

Issue #2462, opened 2023-11-05
docs:  Some Korean translations are missing.


classbinu:
### On which page do you see this issue?

https://daisyui.com/components/?lang=ko

### Describe the issue

I plan to add Korean translations for the items below.

Korean translations are missing for the following items:

### Title
1. Accordion
2. Bottom navigation
3. Join
4. Loading
5. Browser mockup
6. Toast

### Description
1. Accordion
2. Join
3. Loading
4. Browser

saadeghi:
Thanks for the PR

PR #2465, opened 2023-11-05
docs: #2462 Add Korean translation of component page

classbinu:
https://daisyui.com/components/?lang=ko

Korean translations have been added for the items below.

Basically, Google Translator was used for translation and its appropriateness was verified.

The following principles were applied:

1. Use pure Korean whenever possible.
2. Refer to translation examples from Chinese and Japanese, which are in the same Asian region.
3. If the original English language is widely used idiomatically, the English pronunciation is translated into the Korean pronunciation.

### Title
1. Accordion
2. Bottom navigation
3. Join
4. Loading
5. Browser mockup
6. Toast

### Description
1. Accordion
2. Join
3. Loading
4. Browser


saadeghi:
Thank you 💚

classbinu:
Thanks for merging.🥳

I'm very happy to be able to contribute to Daisy UI. 
(Even though it's just a document right now)
I'll contribute to the Bug Fix someday as well.

PR #2472, opened 2023-11-06
docs: add documentation around proper aria roles to follow APG recomm…

KonnorRogers:
Obviously this is incomplete, I'm not overly familiar with DaisyUI and how everything works, simply showing the markup I would expect following the APG guidelines for tabs. It also omits the JS and simply links to a reference implementation from W3C.

### DIFF
--- a/src/docs/src/routes/(docs)/components/tab/+page.svelte.md
+++ b/src/docs/src/routes/(docs)/components/tab/+page.svelte.md
@@ -28,16 +28,32 @@ data="{[
 ]}"
 />
 
+Note, this example does not implement the roving tabindex for tabs, but an example of with their appropriate
+JavaScript can be found here:
+
+<https://www.w3.org/WAI/ARIA/apg/patterns/tabs/>
+<https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/>
+
 <Component title="Tabs">
-<div class="tabs">
-  <a class="tab">Tab 1</a> 
-  <a class="tab tab-active">Tab 2</a> 
-  <a class="tab">Tab 3</a>
+<div class="tabs" role="tablist">
+  <button id="tab-1" class="tab" role="tab" aria-selected="false" tabindex="-1" aria-controls="tab-panel-1">Tab 1</button>
+  <button id="tab-2" class="tab tab-active" role="tab" aria-selected="true" tabindex="0" aria-controls="tab-panel-2">Tab 2</button>
+  <button id="tab-3" class="tab" role="tab" aria-selected="false" tabindex="-1" aria-controls="tab-panel-3">Tab 3</button>
+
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks for the PR.  
This is the first community PR about accessibility in this project so I really appreciate the help 💚

As you mentioned, it's incomplete so I can't simply merge it like this. I'm sure you wouldn't want to merge a PR like this in your project.  

And one of the misconceptions about using daisyUI is that people compare a daisyUI with a JS component like Ant, MUI, etc. where daisyUI is simply a set of additional CSS class names for Tailwind CSS, And framework specific JS component libraries are able to manipulate DOM, add/remove ARIA attributes on demand or they can manipulate keyboard navigation using JS. daisyUI as a CSS file, can not do that obviously. Instead it's suggested to be used alongside headless libraries that can do that (daisyUI for visuals + Headless UI / Radix, etc for functionality)

Given that daisyUI is only a CSS file, I truly care about accessibility. Themes are constantly being improved to provide a better contrast and readability and components are constantly being improved as much as it is possible with CSS+HTML. If you follow the changelog updates you can see how we shifted to use accessible elements like `<dialog>`, `<details>` for interactivities or how we made sure components are accessible for keyboard navigations (using tabs) and screen readers.

For this PR that I truly appreciate, there are following problems and misconception:

### Problem 1
I understand you're not familiar with daisyUI so with all respect, here's the difference between the W3 example you provided and the example daisyUI needs to provide:

daisyUI provides class names that visually look like a tab.
- It can be `<button>` to change a panel like the example from W3 you provided
- It can be `<a>` links to another page (which is standard, accessible HTML tag)

The "tab widget" from W3 example is described as such:  
>[Tabs With Automatic Activation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/): A tabs widget where tabs are automatically activated and their panel is displayed when they receive focus.
[Tabs With Manual Activation](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/): A tabs widget where users activate a tab and display its panel by pressing Space or Enter.  

That W3 page doesn't provide an example where a list of `<a>` links can be styles as tabs (which is common design practice, can be done using standard HTML `<a>` tags)

### Problem 2
Class names `tab-panel` `tab-panel-active` are not daisyUI class names, They don't do anything here so I don't see a reason for them to exist in the example.

### Problem 3
Since daisyUI is not providing any JS bundle file to change content of the tab based on the button click, I there's no reason to include `tabpanel` divs here since all those divs will be visible by default when there's no JS to handle them.

### Acceptable changes
- `role="tablist"`
- `role="tab"`
- `aria-selected="false"` / `aria-selected="true"`

If you can update the PR, I would merge it.

Let me know if there are any disagreements, any questions or any topics to discuss since I'm willing to

saadeghi:
Aria roles added to the docs for `tab`, `alert`, `modal`

PR #2474, opened 2023-11-07
Docs: Accessibility Note on Radial Progress Bar

jodylecompte:
I reviewed #2472 and have tried to consider the discussion there in the changes made. 

**Changes**
This PR adds an explicit role of 'progressbar' to all JSX and HTML examples.

There is already an alert calling out the fact that the standard progress bar uses a `<progress>` element while the radial progress bar uses a `<div>` to work around browser-side issues, this has been expanded to explain that `<progress>` receives it's role implicitly so when using a `<div>` instead we need to add this role explicitly.

**Rationale**
This is not an exhaustive list of the boxes that I would want to see checked from an accessibility perspective, but as mentioned in referenced issue and in Twitter discussion, because Daisy is pure-CSS, there are some things that cannot be demonstrated effectively without JS implementations that are by-intention withheld.

For example
- Best practice would be that a progress bar be labeled and the `progressbar` role allows for labeling via a `<label>` element or one of the `aria-*` labeling properties, but this would be the scope of the parent component that renders the Daisy progress bar, not the progress bar its self directly so this was not included.
- A loading indicator that is being used for a loading-state for a UI that has not yet rendered should receive `aria-busy`, but again this is a consideration of the parent, not the progress bar its self.

**Going Forward**
I wanted to start off small so we can figure out the answers to the questions around where does the Daisy team want to draw the line in regards to what is demonstrated vs what isn't. 

I think that more information would be helpful like advising people of using an indicator of which tab is currently active/selected like the referenced PR, but this is not the most cut and dry work ever because another library doc site that has partial information repeating commonly available information found elsewhere (ie WCAG documentation) so format and leaning heavily on established standards is important. This is probably a larger conversation, but wanted to go ahead and kick the conversation off.

An alternative approach might be a generalized accessibility page that lists the general thoughts expressed here and in #2472 that Daisy inherently can't be treated the same as a fully-fledged (JavaScript) UI kit and a lot will be implemented in user land. Then each component could link to additional information (ie WCAG/APG) so that users can research the requirements. 

Once scope and expectations a little more defined, I'm happy to begin going through other components and looking for opportunities.

### DIFF
--- a/src/docs/src/routes/(docs)/components/radial-progress/+page.svelte.md
+++ b/src/docs/src/routes/(docs)/components/radial-progress/+page.svelte.md
@@ -15,9 +15,11 @@ layout: components
 
 <Translate text="Radial progress needs `--value` CSS variable to work.<br />To change the size, use `--size` CSS variable which has a default value of `4rem`.<br />To change the thickness, use `--thickness` CSS variable which is 10% of the size by default.<br />" />
 
-<div class="alert alert-info text-sm">
+<div class="alert alert-info text-sm mt-2">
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you 💚

PR #2508, opened 2023-11-13
docs: update html for divider (horizontal)

tmthyadms:
Remove flex-col class as it is a horizontal divider.


saadeghi:
Thank you 🙏

PR #2516, opened 2023-11-13
Fix typo probably FF 121+ to 112+

Tronikelis:
It's probably `112` right?

image


saadeghi:
I wish so, but no 🫠  
121 is the future version of Firefox which is going to support CSS `has()` and that's why the icon is currently red.  
https://caniuse.com/css-has

Tronikelis:
Oh damn

PR #2527, opened 2023-11-15
Add 'slate' theme

sveltarne:
The 'slate' theme has the same colors as the old daisyUI dark theme.

While the new theme is a great dark theme, it would be nice for there to be an option for a theme that's dark, but not that dark.

While halloween, dracula, and buisness are similar, none of them match the prettiness of the old dark theme. It was chosen as the default dark theme back then for a reason, I'd assume. Additionally, this would make upgrading past 2.0 easier, as users don't have to use a browser's dev tool's color picker and make a custom theme just to get it looking like it did before. (Yes, while there are files where they could copy the color values from, the average user probably wouldn't find them, especially a beginner.

### DIFF
--- a/src/theming/themes.js
+++ b/src/theming/themes.js
@@ -116,8 +116,8 @@ module.exports = {
   },
   slate: {
     "color-scheme": "dark",
-    "primary": "#088385",
-    "secondary": "#cc4304",
+    "primary": "#661AE6",
+    "secondary": "#D926AA",
     "accent": "#1FB2A5",
     "neutral": "#191D24",
     "neutral-content": "#A6ADBB",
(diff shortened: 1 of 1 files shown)


saadeghi:
What do you think about the 'dim' theme, added with daisyUI 4 release?  
It's not that dark, it has a tone similar to the old `dark` them, however it has different colors for primary, secondary, accent.

sveltarne:
I tried it out as soon as it came out, and while its a nice theme, its alot lighter than the old dark theme if you really take a look, as demonstrated by this image:

image

saadeghi:
Thanks for the PR but another theme so similar to the existing themes is too much,  
I adjusted the background colors of the new `dim` theme to be the same as the old `dark`, It keeps different brand colors tho, because the beuilt-in `light` and `dark` have similar colors already.

PR #2539, opened 2023-11-16
perf: cache functions that get called frequently

gurgunday:
Hey!

Again, great release!

Since we're using the DaisyUI's custom theme generator in our application, performance really matters to us

After a first look, I see that `toGamut` function is always constructed with oklch, so we can extract it to make it a lot faster as V8 will now be able to optimize it, same goes with `cut`

Hope to create more PRs improving performance soon 🙏

### DIFF
--- a/src/theming/functions.js
+++ b/src/theming/functions.js
@@ -3,6 +3,9 @@ const themeDefaults = require("./themeDefaults")
 
 const { toGamut, interpolate, wcagContrast } = require("culori")
 
+const cutNumber = (number) => (number ? +number.toFixed(6) : 0);
+const toGamutOKLCH = toGamut("oklch");
+
 module.exports = {
   isDark: (color) => {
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks for the PR!  
I didn't know wrapping the imported function like this would improve the performance.  
Are there any resources about it where I can learn more?

gurgunday:
Essentially constructing functions inside functions while you can extract them is nearly always a bad practice, as at the end of their execution, they get garbage collected (like other objects and variables) and won’t be optimized

Some say V8 might detect these cases, but at best it’s close to extracting it, and at worst, you really hurt performance

A quick search gave this: https://stackoverflow.com/a/17235220

Also here’s an eslint rule that explains the situation: https://github.com/sindresorhus/eslint-plugin-unicorn/blob/main/docs/rules/consistent-function-scoping.md

saadeghi:
Thank you

PR #2542, opened 2023-11-16
Prefer hasOwn

gurgunday:
We can use hasOwn here as it's safer and [is intended as the modern replacement to `hasOwnProperty`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn)


gurgunday:
Not related to performance this one

PR #2548, opened 2023-11-19
Update docs .env

avirajkhare00:
Added YOUTUBE_API_KEY


saadeghi:
This file resets by GitHub actions on each release.  
You can add a `.env.example` file as a guide and you can put your own API key in `.env.local` file locally (which is ignored by git)

PR #2574, opened 2023-11-25
fix: #2573 tab styles

### LINKED ISSUE #2573: bug: Tab styles are not applied when tabs are set in a hierarchy.

v4.4.6

_No response_

### Reproduction URL

https://play.tailwindcss.com/3QwAlWnlqH

### Describe your issue

Tab styles are not applied when tabs are set in a hierarchy.

yoshi6jp:
fix #2573


saadeghi:
Thank you 🙏

PR #2581, opened 2023-11-26
Update Angular logo

hrstmr:
update logo to use the latest one based on the Angular's press kit

https://angular.dev/press-kit#press-and-media


saadeghi:
Thank you

PR #2586, opened 2023-11-27
I18n: Add Korean translation 

bluejoyq:
## Description
Add Korean translation.

#### before
footer-before

#### after
footer-after

#### before
timeline-before

#### after
timeline-after

#### before
skeleton-before

#### after
skeleton-after

#### before
diff-before

#### after
diff-after


NikarashiHatsu (commented):
Is this PR is still in progress, or does it has been completed? There are still some of untranslated strings on the file.

saadeghi:
Thank you

@NikarashiHatsu I added a batch of missing string to all lang files to make it easier to translate for each language. It doesn't need to be translated 100%. It can be completed progressively.

PR #2596, opened 2023-11-29
Add `.dropdown[open]` data options

koddr:
Hi,

Would be cool to add built-in `daisyUI` option to switch text of the dropdown button when using [`<details>` method](https://daisyui.com/components/dropdown/#method-1-using-details-and-summary-tags).

For example, I always do something like this to solve this:

```diff
<details class="dropdown">
-  <summary class="m-1 btn">open or close</summary>
+  <summary class="m-1 btn" data-opened-text="Close" data-closed-text="Open"></summary>
  <ul class="p-2 shadow menu dropdown-content z-[1] bg-base-100 rounded-box w-52">
    <li><a>Item 1</a></li>
    <li><a>Item 2</a></li>
  </ul>
</details>
```

And CSS:

```css
.dropdown[open] summary::after {
  content: attr(data-opened-text);
}

.dropdown:not([open]) summary::after {
  content: attr(data-closed-text);
}
```

Main motivation: to reduce the amount of custom CSS in projects.

### DIFF
--- a/src/components/styled/dropdown.css
+++ b/src/components/styled/dropdown.css
@@ -18,6 +18,12 @@
 .dropdown:focus-within .dropdown-content {
   @apply scale-100;
 }
+.dropdown[open] summary::after {
+  content: attr(data-close);
+}
+.dropdown:not([open]) summary::after {
+  content: attr(data-open);
+}
 @media (hover: hover) {
   .dropdown.dropdown-hover:hover .dropdown-content {
     @apply scale-100;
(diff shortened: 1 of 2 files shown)


saadeghi:
This would create accessibility issues.  
`<summary>` tag must have a text or aria label.   
You can do something like this to avoid custom CSS:  
https://play.tailwindcss.com/JMnlWpCUQ5
```html
<details class="dropdown">
  <summary class="btn m-1">
    <span class="[[open]>*>&]:hidden">Open</span>
    <span class="hidden [[open]>*>&]:inline">Close</span>
  </summary>
  <ul class="dropdown-content menu z-[1] w-52 rounded-box bg-base-100 p-2 shadow">
    <li><a>Item 1</a></li>
    <li><a>Item 2</a></li>
  </ul>
</details>
```

Issue #2622, opened 2023-12-06
bug: drawer-end makes horizontal scrollbar permanently visible.


kobe-ra:
### What version of daisyUI are you using?

v4.4.19

### Which browsers are you seeing the problem on?

Chrome

### Reproduction URL

https://daisyui.com/components/drawer/#drawer-that-opens-from-right-side-of-page

### Describe your issue

Horizontal scrollbar is permanently visible when using `drawer drawer-end`.
This issue is also visible in the docs.

![image](https://github.com/saadeghi/daisyui/assets/32091524/649b71d4-51a5-47f5-a0b4-f92ff41a3db3)

l1c2d3:
I'm observing the same behavior in Edge. Firefox renders correctly.
I'm using Next.js 14, tailwindcss 3.3.6, daisyui 4.4.19

I'm using drawer drawer-end in a root layout and that's when this happens. If I change drawer drawer-end to a template or page, the scroll disappears, but the drawer slides from the left, instead of right.

Also, I'm observing some styles not rendering correctly in Edge (oversized components) but correct in Firefox.
This was working fine last week, so this bug must have been introduced in the most recent versions.

I will test with previous releases.

gagansuie:
hi! solved the issue by adding overflow-x-hidden tailwind class to <div class="drawer-side"

LanceMoe:
> hi! solved the issue by adding `overflow-x-hidden` tailwind class to `<div class="drawer-side"`

In my project, by adding `overflow-x-hidden` can be solved on the desktop, but on the mobile phone, the entire screen can be swiped left or right through the touch screen.

After returning to 4.4.14, this problem was solved.

gagansuie:
If that's the case, you can do md:overflow-x-hidden or something to do this fix on certain screen sizes.

gereltod-g:
Using version 4.4.19, this error is still occurring. It's not noticeable on macOS Chrome, but on Windows, an unscrollable scrollbar appears

bayasdev:
> md:overflow-x-hidden

+1

PR #2630, opened 2023-12-08
fix: #2622 drawer-end makes horizontal scrollbar permanently visible

binjuhor:
+ Updated src/components/styled/drawer.css to add overflow-hidden


src/components/styled/drawer.css
```diff
@@ -1,6 +1,7 @@
 .drawer {
   width: 100%;
   &-side {
+    @apply overflow-hidden;
```
saadeghi, line 4: overflow hidden is not the solution. if sidebar content is long, it should overflow.

saadeghi (changes requested):
What is the problem? can you give a screenshot?

src/components/styled/drawer.css
binjuhor: This is the problem .
image

>overflow hidden is not the solution. if sidebar content is long, it should overflow. 

yeah I see, so my solution is `@apply overflow-x-hidden;`  @saadeghi

lucasrivoiro:
I have same issue here, but applying overflow-x-hidden to drawer-side create another issue for me. Applying overflow-x-hidden turns the entire window vertically scrollable, but horizontally fix the issue.

binjuhor:
@lucasrivoiro Can you share me the screenshot? I think the window vertical should scrollable because the content can be so long like this

https://github.com/saadeghi/daisyui/assets/4584599/15982d8e-3d5a-44cb-ba45-f069f040ebc7

LanceMoe:
At Windows, Or set the scroll bar to always show in macOS, the Drawer will show a scroll bar.
Before version 4.4.14, it worked well.
image

brendoarrudaa:
Any chance of error being adjusted? I'm in a project that after upgrading from version 3 to 4 broke the entire drawer end of the page

cloudhoy:
@brendoarrudaa for now, you can fix this by adding `overflow-x-hidden`  to element with the 'drawer-side' class:
```
<div class="drawer-side overflow-x-hidden">
  {/* Sidebar content here */}
</div>
```

binjuhor:
@cloudhoy Yeah that also my solution, but I don't know why this PR has not merged yet.

brendoarrudaa:
@cloudhoy  I share the same feeling, it just needs to be approved to resolve the bugs,
@saadeghi Any chance of being approved?

PR #2635, opened 2023-12-08
fix: #2634 compile base colors

### LINKED ISSUE #2634: bug: Not compiling `theme(colors.<x>)` in `/base`

v4.4.19

Other

### Reproduction URL

https://stackblitz.com/edit/daisyui-vite-l1jv1a?file=package.json

### Describe your issue

You can see the issue in the [stackblitz repro](https://stackblitz.com/edit/daisyui-vite-l1jv1a?file=package.json) by running the terminal:

```sh
cat node_modules/daisyui/dist/base.js
```

```js
// actual:                                                   🚨                               🚨
module.exports = {":root,\n[data-theme]":{"backgroundColor":"theme(colors.base-100)","color":"theme(colors.base-content)"} // ... ...
// expected:
module.exports = {":root,\n[data-theme]":{"backgroundColor":"var(--fallback-b1,oklch(var(--b1)/1))","color":"var(--fallback-bc,oklch(var(--bc)/1))"} // ... ...
```

It seems that the file is being skipped in the compilation step. Will investigate further.

gorango:
Transform theme colors in the base colors configuration.

Resolves #2634

### DIFF
--- a/src/base/postcss.config.js
+++ b/src/base/postcss.config.js
@@ -2,6 +2,7 @@ module.exports = {
   plugins: [
     require("postcss-import"),
     require("tailwindcss/nesting"),
+    require("tailwindcss")("./src/components/tailwind.config.js"),
     // require("autoprefixer"),
   ],
 }
--- a/src/base/postcss.config.js
+++ b/src/base/postcss.config.js
@@ -2,7 +2,7 @@ module.exports = {
   plugins: [
     require("postcss-import"),
     require("tailwindcss/nesting"),
-    require("tailwindcss")("./src/components/tailwind.config.js"),
+    require("tailwindcss")("./src/base/tailwind.config.js"),
     // require("autoprefixer"),
   ],
 }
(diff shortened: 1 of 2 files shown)


src/base/postcss.config.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   plugins: [
     require("postcss-import"),
     require("tailwindcss/nesting"),
+    require("tailwindcss")("./src/base/tailwind.config.js"),
```
saadeghi, line 5: file `./src/base/tailwind.config.js` doesn't exist.
Can you add it with following content?
```
module.exports = {
  theme: {
    colors: require("../theming"),
  },
}
```

saadeghi (changes requested):

gorango:
Good catch - my bad.

src/base/postcss.config.js
gorango: Good catch - my bad

Issue #2651, opened 2023-12-13
bug: Can't Override Input Size Attributes


gblandinkingland:
### What version of daisyUI are you using?

v4.4.19

### Which browsers are you seeing the problem on?

Edge

### Reproduction URL

https://play.tailwindcss.com/ftoGaT3hDA

### Describe your issue

The padding and height attributes of inputs cannot be overridden. This is likely the same issue from #2499 just in [/src/utilities/unstyled/input.css](https://github.com/saadeghi/daisyui/blob/master/src/utilities/unstyled/input.css) instead of [/src/components/styled/input.css](https://github.com/saadeghi/daisyui/blob/master/src/components/styled/input.css).

The following code snippet won't work in v4 but did in v3.
```<input class="input input-sm input-primary pl-8 m-5"></input>```
Of course, similar to in the mentioned issue above, adding important makes this work correctly.
```<input class="input input-sm input-primary !pl-8 m-5"></input>```

The URL provided has this exact example to demonstrate.

PR #2652, opened 2023-12-14
[Bug fix] Fix the issue "Can't Override Input Size Attributes"

souyang:
Fix the bug https://github.com/saadeghi/daisyui/issues/2651

Steps to reproduce: 
1. Checkout the branch
2. Run `npm install && npm run build && npm run playground`
3. In `src/playground/components/Component.astro`, update content to
```html 
<div>
  <input class="input input-sm input-primary m-10"></input>
</div>
```
Then update content to
```html
<div>
  <input class="input input-sm input-primary pl-8 m-10"></input>
</div>
```

**Expected result**: the left padding should be updated to 8 px
**Actual result**: the left padding is unchanged.

Testing Steps: 
1. Run `npm install && npm run build && npm playground`
2. In `src/playground/components/Component.astro`, update content to
```html 
<div>
  <input class="input input-sm input-primary m-10"></input>
</div>
```
Then update content to
```html
<div>
  <input class="input input-sm input-primary pl-8 m-10"></input>
</div>
```
**Expected result**: the left padding should be updated to 8 px
**Actula result**: the left padding was updated to 8 px


src/utilities/unstyled/input.css
```diff
@@ -1,5 +1,4 @@
-.input,
-input.input {
+input {
```
saadeghi, line 1: It should be `.input` not `input`

saadeghi (changes requested):

src/utilities/unstyled/input.css
souyang: Comment is resolved. Sorry for the typo

Issue #2638, opened 2023-12-09
bug: style broken at japanese documentation


Kenzo-Wada:
### What version of daisyUI are you using?

latest

### Which browsers are you seeing the problem on?

Chrome

### Reproduction URL

https://daisyui.com/?lang=ja

### Describe your issue

![IMG_2429](https://github.com/saadeghi/daisyui/assets/79452224/409891bb-60bf-42ab-899d-34e73b8c0ea6)

some components is displayed outside of navbar. I think this is because component is to long. I suggest to show only japanese component name not both english and japanese to fix this issue (as other languages does)

If it is welcome, I'm happy to submit PR:)

saadeghi:
I see.  
The problem (according to translators) is that the name of some components may be unrecognizable if we translate them to native language because they are technical terms (kind of)  
So the solution translators suggested was to include both English and Japanese names (which is too long together)

Which way do you think is more usable?
- English names for all components?
- Japanese translation for all components?
- Translating the ones that can have a understandable translation + not translating the ones that would feel weird if translated?

Kenzo-Wada:
I see.

> Japanese translation for all components?
though it is small deference between those three, I prefer this one, (because other languages does so)

saadeghi:
Can you please check all the item names on the sidebar and tell if any of the component names on the sidebar would sound unclear/weird/confusing if we only use translated word?

Kenzo-Wada:
@saadeghi 

Sorry for taking time.

some component name which uses "Kanji" feeled weird for me. ex.)https://daisyui.com/components/progress/ 
There sounded more likely to Chinese.

Should I correct these component name and submit PR?

saadeghi:
>Should I correct these component name and submit PR?

Yes that would be great!  

It's not important that everything is translated. Especially technical words or the words that their English string is more informative.

So for example if the word `progress` has a Japanese translation that is accepted among Japanese developers and designers, we can use the Japanese translation. Otherwise we can use the English string.

PR #2666, opened 2023-12-15
Fix Navbar style broken at JA page

Kenzo-Wada:
Closes: https://github.com/saadeghi/daisyui/issues/2638

After careful consideration, I found that the English component names were easier to understand, so I modified them to English only.
There are two reasons for this
- Component names are almost like proper nouns, so they are easier to understand if they are not badly translated.
- The English used in component names is often simple, so it is likely that many people can understand them.

In addition, we also corrected the accompanying translations that were clearly incongruous. There is still room for improvement, but may we continue to work on correcting the Japanese translation at another PR?


saadeghi:
>may we continue to work on correcting the Japanese translation at another PR?

Sure!

Thank you so much.

PR #2672, opened 2023-12-17
Update tailwind.config.js

GAUTAMMANAK1:
add another color theme


src/components/tailwind.config.js
saadeghi, line 1: What is this?

PR #2677, opened 2023-12-18
Updating theming docs

TAQ2:
When I tried what was in the docs it didn't work. Found the solution on stack overflow


saadeghi:
This is docs for daisyUI 4.  
If you're using daisyUI 3, use [version 3 docs](https://v3.daisyui.com/docs/themes/#-7)

PR #2701, opened 2023-12-27
When the language is Chinese, there is a problem with the File Input and Theme controller display in the menu bar.

dearDreamWeb:
When the language is Chinese, there is a problem with the File Input and Theme controller display in the menu bar.

### DIFF
--- a/src/docs/src/lib/data/pages.js
+++ b/src/docs/src/lib/data/pages.js
@@ -85,7 +85,7 @@ export const pages = [
             href: "/components/swap/",
           },
           {
-            name: "Theme Controller",
+            name: "Theme controller",
             tags: "theme controller change dark mode",
             href: "/components/theme-controller/",
             badge: "new",
@@ -264,7 +264,7 @@ export const pages = [
             href: "/components/checkbox/",
           },
           {
-            name: "File Input",
+            name: "File input",
             tags: "file upload",
             href: "/components/file-input/",
           },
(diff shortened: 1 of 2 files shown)


saadeghi (commented):
What is the problem? and why change them to lowercase?

dearDreamWeb:
> What is the problem? and why change them to lowercase?
Previously, the Chinese translation of the left navigation bar was English plus Chinese. File Input had this copywriting before, which would cause the copywriting of the left navigation bar to be only Chinese, not English plus Chinese. So refer to the Text input to modify it.

saadeghi:
Is it required to include both Chinese and English names?  
We had this issue with Japanese translation as well, the problem is, contributors translated all the strings to native language but most names were uninformative because component names are technical design/development terms. So then another contributor appended english strings to fix the issue but that doesn't look good in my opinion.

As a Chinese speaker, do you think all/some names for component names on sidebar is good enough or do they need Chinese translation?

dearDreamWeb:
For new developers, Chinese translation is also needed

saadeghi:
But, is the English string needed as well? For all component names? I'm looking for a way to use one of them and I think if a translated component name is informative enough, it's safe to translate it. If the translation makes it confusing, we can only use English name. But using them both at the same time, next to each other, defeats the whole purpose of translation.

dearDreamWeb:
It should be a Chinese habit to use English strings. You can take a look at antd's component library (https://ant.design/components/overview-cn/). Of course, it is also possible not to use English strings. After all, people from other countries may also use daisyui. As a Chinese, it may take some time for me to adapt.

saadeghi:
So if possible, in a new PR, please use either English or Chinese for component names.  
Using both doesn't seem to be a good solution and of course having lower case and caption case keys for the same string is not a good idea.

PR #2705, opened 2023-12-27
Fix docs about default size of Radial Progress

MMore:
[Here](https://github.com/saadeghi/daisyui/blob/master/src/components/styled/radial-progress.css) you find the configured default.


saadeghi:
Thanks

PR #2708, opened 2023-12-28
added translation function to component component

dearDreamWeb:
I added a translation function to the component component, and now I have finished translating the Button related stuff in the component.


saadeghi:
Thank you!

Issue #2110, opened 2023-07-08
docs: navbar example does not match actual style


debashisbiswas:
### On which page do you see this issue?

https://daisyui.com/components/navbar/

### Describe the issue

In the HTML example code for each navbar, each `div` is styled with `class="navbar bg-base-100"`. However, on the Preview tab, this is not what is shown; the previewed navbar is styled with `class="navbar bg-base-100 shadow-xl rounded-box"`.

### Example code:
![image](https://github.com/saadeghi/daisyui/assets/37234687/df1bfc0a-bf1f-4389-9062-665dae84da3b)

### Actual navbar:
![image](https://github.com/saadeghi/daisyui/assets/37234687/8cb89b22-a2bc-499b-80ba-22c619d081bc)

### What browsers are you seeing the problem on?

All browsers

saadeghi:
It's intentional.  
When people copy/paste the code, most of them need a navbar with no border radius and they want to stick it to the top of their content because that's a more common design pattern.  

What you see on the site as the demo has a shadow and border radius to make it visible and distinguishable from the background. Otherwise it would be hard to see if it has no shadow or border.

And after all, shadow and border radius are Tailwind CSS utility classes and any one can use them if they want.

perelin:
Just also stumbled over this. @saadeghi Your reasoning is valid, but the code examples missing border radius is very confusing. I am currently trying to get familiar with all available components and was searching over 1h for any mistakes I made that could cause the missing border radius. It should be pointed out in the documentation. Let me see if I can create a pull request for you.

PR #2709, opened 2023-12-28
Added information about missing classes in navbar code examples

perelin:
See also https://github.com/saadeghi/daisyui/issues/2110#issuecomment-1627252219


saadeghi (changes requested):
You need to import `Translate` before using it
```js
import Translate from "$components/Translate.svelte"
```

PR #2714, opened 2023-12-29
docs: updated Polish translation

Avaray:
In Polish language the word "Library" is of feminine gender. That's why it was necessary to change a few words.
Additionally, the word "Welcomes" (line 386) is challenging to translate literally. I decided that expressing thanks for the developers' contributions would be the best choice in this case.


saadeghi:
Thank you!

Issue #2690, opened 2023-12-21
bug: No border above table footer


jlous:
### What version of daisyUI are you using?

4.4.23

### Which browsers are you seeing the problem on?

Chrome

### Reproduction URL

https://play.tailwindcss.com/VV44TFAzwW
https://daisyui.com/components/table/#table-with-visual-elements

### Describe your issue

In a table with a single footer row and the basic 'table' style (with hairlines between rows), I would expect a line between the footer and the body.
This would mirror the header, and follow the general look of the basic table.

In stead, the footer  and the last row of the body look like they are somehow grouped as a single row.

Nekxio:
Hello 😄
I'll take a look !
Which branch should I use to fix this issue ? @saadeghi

Issue #2692, opened 2023-12-21
bug: Multi-row table headers get weird borders


jlous:
### What version of daisyUI are you using?

4.4.23

### Which browsers are you seeing the problem on?

Chrome

### Reproduction URL

https://play.tailwindcss.com/exLChgw4il

### Describe your issue

If my table header has several rows, the border styles get very weird:  

1. The border between the header and body disappears - the most significant split in the whole table
2. Borders appear between the header rows, which you don't really want

Even if point 1 was solved, separating the header rows with lines only makes it less clear where the header stops and the body starts.

PR #2723, opened 2024-01-03
fix: #2690 Add border above table footer

Nekxio:


Nekxio:
I noticed that issue #2692  concerned the same code, so I took the liberty of adding a new commit to this PR 😄

saadeghi:
Thanks!

Issue #2571, opened 2023-11-24
bug: Dropdown affects height while hidden


eulertour:
### What version of daisyUI are you using?

v3.3.5

### Which browsers are you seeing the problem on?

Chrome

### Reproduction URL

https://play.tailwindcss.com/7X98V0vXQg

### Describe your issue

The height of a dropdown created using [method #2 from the documentation on dropdown](https://daisyui.com/components/dropdown/#method-2-using-label-and-css-focus) can overflow the parent elements, which breaks scrolling. This might be the same problem encountered in #1276.

Nekxio:
Hello @eulertour 😄 
If I understood your initial request correctly, I think the problem came from this line of code :
`<label tabindex="0" class="btn m-1">open/close dropdown</summary>`

By replacing the correct tag closure, it works. I've also tested replacing label with summary and it's the same result.

eulertour:
I'm still seeing the problem after matching either <label> or <summary> tags. I was trying to prevent the collapsed menu from increasing the scroll height.

Nekxio:
I understand better after retesting. I'll look into it right away !

darklow:
Could this please be merged? I confirm bug is still present and adding dropdown stretches parent container because "visibility" is used not hide/block approach.  Thank you

saadeghi:
@darklow I understand but this is a breaking change, any site that is relying on the current visibility and opacity will break so I can't change it in a patch version or minor version. It must be major version.  
We should wait for daisyUI 5. Until then, add some custom styles to get the desired result.

PR #2736, opened 2024-01-07
fix: #2571 Fix dropdown that increased page size

Nekxio:


saadeghi:
Thanks for the PR but I can not merge this.

This removes the animation.  
And more importantly it may break existing websites that are relying on the current styles.

Issue #2775, opened 2024-01-19
bug: text opacity on disabled select has no effect


skayo:
### What version of daisyUI are you using?

v4.6.0

### Which browsers are you seeing the problem on?

Firefox

### Reproduction URL

https://play.tailwindcss.com/EFs2R8UIwp

### Describe your issue

The select component has a `text-opacity-20` modifier when it is disabled, but due to there being no `text-base-content` (or similar) on the select element, the opacity modifier has no effect.

Here is the line where `text-base-content` (or similar) is missing: https://github.com/saadeghi/daisyui/blob/214cc01260ef79cea584f249d8ef830429fe2690/src/components/styled/select.css#L70

Nekxio:
Hello @skayo,
Indeed, there was a problem with the disabled select. I took the liberty of making a PR to resolve the issue. Thanks for your help 🙂

PR #2776, opened 2024-01-20
fix: #2775 Fix disabled select opacity modifier has no effect

Nekxio:


saadeghi (changes requested):
Please only commit the related CSS file.

Nekxio:
It's done 😄

Issue #2625, opened 2023-12-07
bug: The scroll button is at the middle of the input-sm component 


hughlv:
### What version of daisyUI are you using?

v4.4.14

### Which browsers are you seeing the problem on?

Edge

### Reproduction URL

https://daisyui.com/components/input/

### Describe your issue

Please see this screenshot:

<img width="59" alt="image" src="https://github.com/saadeghi/daisyui/assets/817110/a96d3beb-4115-4dd7-88ff-ce8abbbeaaa3">

The problem is that padding-right is set 0.75rem. It's fine for normal text input, but makes the scroll up/down buttons positioned in a weired place.

<img width="199" alt="image" src="https://github.com/saadeghi/daisyui/assets/817110/5aa5d135-263e-4a3b-8778-7976efbd959d">

If I change the padding-right to 0 mannually then the scroll buttons will be at the right place.

PR #2777, opened 2024-01-20
fix: #2625 Fix input spin button position

Nekxio:


saadeghi (changes requested):
This doesn't fix the issue. The margin is already `0`  
Something like this can fix it:
```css
.input[type="number"]::-webkit-inner-spin-button,
.input-md[type="number"]::-webkit-inner-spin-button{
  @apply -my-4 -me-4 cursor-pointer;
}
.input-xs[type="number"]::-webkit-inner-spin-button{
  @apply -my-1 -me-2 cursor-pointer;
}
.input-sm[type="number"]::-webkit-inner-spin-button{
  @apply my-0 -me-3 cursor-pointer;
}
.input-lg[type="number"]::-webkit-inner-spin-button{
  @apply -my-6 -me-6 cursor-pointer;
}
```
The negative margin covers the padding. Also each size has different paddings.

Nekxio:
Indeed, your proposal works.
I've just changed the values of input-xs and input-sm to get the desired rendering.

PR #2791, opened 2024-01-24
Use h6 instead of header in footer component examples

### LINKED ISSUE #2789: bug: header elements cannot be placed inside a footer element

4.6.0

All browsers

### Reproduction URL

https://daisyui.com/components/footer/

### Describe your issue

Hello, your footer examples need a small fix, as specified by w3c (https://www.w3schools.com/tags/tag_header.asp) and mdn (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/header#technical_summary) header elements cannot be placed inside a footer element, should consider using divs?

_Originally posted by @bempensato in https://github.com/saadeghi/daisyui/issues/2294#issuecomment-1908788727_

danawoodman:
As @bempensato mentioned in #2294 (and reported in #2789), having a `<header>` as a child of `<footer>` is invalid.

Moving to use an `<h6>` tag as it has the lowest weight of a page's heading tags and still properly represents a heading of the `<nav>` section.

Ran through the W3C validator to confirm, passes with flying colors:
Showing results for contents of text-input area - Nu Html Checker 2024-01-24 at 3 44 44 PM

Note I also changed `<br/>` to `<br>` as the validator warns the trailing slash is unnecessary.

Closes #2789

### DIFF
--- a/src/docs/src/routes/(docs)/components/footer/+page.svelte.md
+++ b/src/docs/src/routes/(docs)/components/footer/+page.svelte.md
@@ -27,21 +27,21 @@ data="{[
 <Component title="Footer">
 <footer class="p-10 footer bg-neutral text-neutral-content rounded">
   <nav>
-    <header class="footer-title">Services</header> 
+    <h6 class="footer-title">Services</h6> 
     <button class="link link-hover">Branding</button>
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you @danawoodman

PR #2799, opened 2024-01-28
Add new buttons with ring

peymanmotamedifar:


saadeghi:
Thank you for the PR.  
Can you please let me know what's the use case for this?

peymanmotamedifar:
> Thank you for the PR. Can you please let me know what's the use case for this?

First of all I am glad to start a communication with you sir, and about my commit, I am a daysiui fan in my company so I added a new decoration type for the buttons of daisyui. I decided to named my design "Buttons with ring". I would be interested in doing more design for daisyui materials so if there is some issues for work you and your team can count  on me.

Thank you for contacting me.

saadeghi:
Thank you for the effort but I wish you [opened an issue about it before making a PR](https://github.com/saadeghi/daisyui/blob/master/.github/CONTRIBUTING.md) to make sure it would be something we want to have in the library.  

For buttons, we're using `outline` for focus within style and the hover style is about changing the background color slightly. It is possible to apply custom styles using Tailwind CSS utility classes but I'm not sure if having this style would add a value to the library.

I appreciate the effort and I would welcome any PR regarding to fix an [existing issue](https://github.com/saadeghi/daisyui/issues). To start fixing an issue please ask me to assign the issue to you so we can avoid duplicate or unwanted effort ([read more](https://github.com/saadeghi/daisyui/blob/master/.github/CONTRIBUTING.md#fixing-existing-issues)) 💚

Issue #2788, opened 2024-01-24
bug: Hero with overlay image - bg-opacity-[number] is not working properly on Android phone


duc-hoang-161:
### What version of daisyUI are you using?

4.6.0

### Which browsers are you seeing the problem on?

Chrome Android

### Reproduction URL

https://daisyui.com/components/hero/#hero-with-overlay-image

### Describe your issue

Chrome Android: v103
Root cause: oklch is not supported. [oklch doc](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch)
Suggestion: Use opacity-[number] instead of bg-opacity-[number]

I'm using the component Hero with overlay image in my project and noticing a weird bug on an Android phone when the overlay seems solid even though it can see through on other browsers (Desktop, iPhone). I made a quick fix on the browser by applying opacity to the overlay element and it worked.

![Screenshot 2024-01-24 at 10 18 38 AM](https://github.com/saadeghi/daisyui/assets/31062901/413bb582-3843-4702-9059-de19073895cd)
![Screenshot 2024-01-24 at 10 19 30 AM](https://github.com/saadeghi/daisyui/assets/31062901/679ac54b-7330-456a-865a-3ea54a9c8186)

duc-hoang-161:
A related issue https://github.com/saadeghi/daisyui/issues/2703

PR #2815, opened 2024-02-03
fix: #2788 fix the example of using opacity on hero overlay to work on old browsers

duc-hoang-161:
I simply replaced `bg-opacity-[number]` with `opacity-[number]` to make it works on old browsers due to the lack of supporting `oklch` ([oklch doc](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/oklch)) on old browers.


saadeghi:
Thanks for the PR.

That won't look the same because `.hero-overlay` has a background opacity of `0.5` and making the whole div transparent would look really bright (it will be 0.5 bg opacity and 0.6 opacity at the same time)

I would suggest updating the browser or using daisyUI 3.x

PR #2844, opened 2024-02-13
Update german translations

Flo0807:
I am a native German speaker and I noticed that most of the German translations are not of good quality. So I completely reworked them.

Thanks for your good work. Keep it up!


saadeghi:
Thank you!

PR #2853, opened 2024-02-17
Fixed flawed example code for Accordion

Supermarcel10:
Fixed an issue with the example code for accordion.

The flawed code:
- Rendered a read-only field due to checked being set to a value instead of using "defaultChecked".
- Raised an error in TypeScript (TS2322: Type `string` is not assignable to type `boolean | undefined`)

This addresses both those issues

### DIFF
--- a/src/docs/src/routes/(docs)/components/accordion/+page.svelte.md
+++ b/src/docs/src/routes/(docs)/components/accordion/+page.svelte.md
@@ -41,7 +41,7 @@ data="{[
 
 <Component title="Accordion using radio inputs">
 <div class="collapse bg-base-200">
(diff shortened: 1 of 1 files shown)


saadeghi (changes requested):
`defaultChecked` is **not** a standard HTML attribute. It's JSX.  
To fix the docs example for JSX, please update the function that converts HTML to JSX instead: 
- Add `'checked="checked"': "defaultChecked",` to https://github.com/saadeghi/daisyui/blob/master/src/docs/src/lib/actions.js#L28

Supermarcel10 (commented):
Original changes have been reverted. Instead `actions.js` has been used to  replace the string as requested.

saadeghi:
Thank you

Issue #2832, opened 2024-02-07
docs: range with custom color


iamsatyajit05:
### On which page do you see this issue?

https://daisyui.com/components/range/

### Describe the issue

There is no mention of how to add custom color to the range. A mention of it saves a little bit of time for most developers. Because the same thing is mentioned for the toggle. 

It can be done with [--range-shdw:colorname], that can be directname, #hexvalur, rgb(values)

### What browsers are you seeing the problem on?

All browsers

iamsatyajit05:
@abhishekkr (sorry for disturb)

abhishekkr:
@iamsatyajit05 `--range-shdw` is (just) mentioned at [utilities](https://daisyui.com/docs/utilities/#-4)

It would take me sometime to get to this... if you can raise a PR for the docs, I guess as soon @saadeghi or other maintainers get some time... it would be a quick merge.

iamsatyajit05:
Okay, thanks.

PR #2866, opened 2024-02-23
docs: #2832 range with custom color

iamsatyajit05:
There is no mention of how to add custom color to the range. A mention of it saves a little bit of time for most developers. Because the same thing is mentioned for the toggle.


saadeghi:
Thanks

PR #2889, opened 2024-03-06
Add carousel-start modifier

2hu12:
If we add `carousel-center` to the classname for mobile devices and it'd be hard to change it back to the default `scroll-snap-align: start` behaviour for desktop. So just add this `carousel-start` modifier and use `lg:carousel-start` for this senario.


saadeghi:
Thank you

PR #2901, opened 2024-03-11
docs: Add another WordPress installation

suabahasa:
Add an example of how to set up daisyUI for WordPress using the [Yabe Siul](https://siul.yabe.land/) plugin


saadeghi:
Thank you for the PR.  
I apologize but this page can only include links to free tools/libraries.

PR #2933, opened 2024-03-24
add: added Bengali translation

mdsabbiralmamon:
I have added the Bengali translation file and translated almost 99% of translation. I have seen in discord server a few asking about it so you can add and marge this also if there is any issues let me know I will fix it as soon as possible.


saadeghi:
Thanks!

PR #2963, opened 2024-04-08
Docs: Update theme generator

TreyRuffy:
Added a button to show all colors in the theme generator.

With the switch off:
image

With the switch on:
image

### DIFF
--- a/src/docs/src/routes/(docs)/theme-generator/+page.svelte
+++ b/src/docs/src/routes/(docs)/theme-generator/+page.svelte
@@ -88,7 +88,7 @@
     "error",
   ]
 
-  $: onlyRequiredColorNames = false
+  $: onlyRequiredColorNames = true
 
   $: colors = [
     {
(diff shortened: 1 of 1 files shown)


TreyRuffy:
I realized I have to make a couple more changes before it can be merged. Expect them later today / tomorrow.

TreyRuffy:
I forgot to leave a comment but this should now be ready for reviewing. Lmk if there's anything I should change!

saadeghi:
Hey,  
Thanks for the PR, (and sorry for the spaghetti code of this page 😅)  
There's an issue with the value of all optional colors. It shows a color on the demo but another color on the code:  
Screenshot 2024-04-13 at 3 58 54 PM

You can see this happening by pressing the random button a few times. For example in above screenshot, you can see `*-content` color of buttons don't match the ones in the code. Or the `base-content` color which should be dark in this screenshot but in the preview is light.

TreyRuffy:
I was able to fix that issue in the latest commit. Additionally, I added a button to automatically generate the optional colors.

Show all enabled:
image

Show all disabled:
image

sovetski:
Looks very good! Maybe it is not the correct place here, but what do you think guys about adding also `border radius` config to the theme generator config?

saadeghi:
Thanks

PR #2996, opened 2024-04-24
Update +page.svelte.md to fix Hero component documentation

Hypnocratic:
Hero component code example of "Hero with figure but reverse order" was wrong. The image tag was being put before the text div. Fix moves image to after text div. Code was causing image to render left of text instead of the right.


saadeghi:
It's not wrong.  
[Second example](https://daisyui.com/components/hero/#hero-with-figure) has `lg:flex-row`  
[Third example](https://daisyui.com/components/hero/#hero-with-figure-but-reverse-order) has `lg:flex-row-reverse`  


You can copy the codes to Tailwind Play and see they're working as expected:  
https://daisyui.com/tailwindplay/

PR #2999, opened 2024-04-25
update install markdown, use ESM for tailwind config.

s0h311:
Tailwind is using ESM for its config file by default for while now.


saadeghi:
> Tailwind is using ESM for its config file by default for while now.

It's commonJS by default: https://tailwindcss.com/docs/installation

We're not going to **replace** the commonJS example, but yeah the ESM example will be added next to that. There's an open PR for that already: https://github.com/saadeghi/daisyui/pull/2989/files

s0h311:
mb, it really is cjs by default.

PR #3000, opened 2024-04-25
Use ESM for tailwind config

flexchar:
I noticed I often have to edit this in my own setup. Since ESM is now common, it would be nice to have module-ready example. :)


saadeghi:
Thanks for the PR

We're following the default installation guide from Tailwind CSS which is commonJS:  
https://tailwindcss.com/docs/installation

I will add a ESM config as well but it's not a good idea to completely remove commonJS and replace it with ESM especially when Tailwind goes by commonJS by default. It would be confusing for people who are using commonJS.

flexchar:
Good point.

Issue #2950, opened 2024-04-01
docs: Inaccessible components / wrong advice on docs.


horsemankukka:
### Describe the issue

Multiple daisyUI components or their example uses are problematic concerning accessibility. You asked me on Twitter to write about them here, so here goes.

## Countdown

Because the countdown component is essentially just a pseudo-element containing numbers from 00 to 99, screen readers will read it as such.

## Timeline

The [first example](https://daisyui.com/components/timeline/) gets read as "1984. First Macintosh computer. Divider. Disabled. Divider. Disabled. 1998. iMac. Divider. Disabled. Divider. Disabled."

Using those two `<hr>`s without hiding either one of those from the screen reader makes reading the timeline tedious.

## Radial progress

The [docs](https://daisyui.com/components/radial-progress/) claim that adding [`role=progressbar`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/roles/progressbar_role) makes the element accessible again. In fact it does the opposite: just declaring the role but not `aria-valuenow` (even though ARIA requires it) makes all of the examples read "50%" on VoiceOver, for example.

## Tabs 

This one misuses the [tablist role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/tablist_role) as it doesn't implement all roles it needs. The `input + div + input + div` order also makes navigating the tabs weird and unexpected. 

## Rating

The examples show just unlabeled radio buttons without any indication to screen reader used about what they do.

## Drawer

The drawer's content isn't hidden for screen reader users ever. Additionally, even though it's redundant as the drawer isn't actually hidden for screen reader users, the `<label for aria-label>` pattern doesn't seem to work in VoiceOver (it doesn't get announced and cannot be selected). MDN docs for [`aria-label`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-label) states that in practice it's supported for limited number of elements only, might be about this. 

## Using checkboxes for non-checkbox stuff

This is just confusing overall, no matter what.

---

Not sure if this covered everything, feel free to mention if there was something I said on Twitter but forgot to include here. Marked as an issue on docs even though it's not purely just that and didn't want to open different issues for every single component.

gjvalenz:
Hey, I'd like to work on this. I'm using NVDA to test which does show some discrepancies from VoiceOver (it has no problems reading Drawer aria-labels, etc).

[...]

PR #3002, opened 2024-04-27
Fixes for issue #2950

gjvalenz:
Please look over to ensure everything is fine! I did my own testing, but it's always better to have multiple pairs of eyes rather than just one. Be warned: It's a big pull!

### DIFF
--- a/src/components/unstyled/countdown.css
+++ b/src/components/unstyled/countdown.css
@@ -3,7 +3,7 @@
 }
 .countdown {
   display: inline-flex;
-  & > * {
+  & > *:not(.countdown-accessible) {
     height: 1em;
     @apply inline-block overflow-y-hidden;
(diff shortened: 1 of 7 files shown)


nandi95:
Do any of these need [`aria-live`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)?

gjvalenz:
> Do any of these need [`aria-live`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions)?

Good question.

I think countdown and radial progress might need them!

saadeghi:
Thanks for the PR 💚  
daisyUI 5 is now available and all the issues are addressed in daisyUI 5 and also in the docs.

Issue #3009, opened 2024-05-01
bug: disabled checkbox border overflow


nandi95:
### What version of daisyUI are you using?

"daisyui": "^4.4.20"

### Which browsers are you seeing the problem on?

Chrome, Firefox

### Reproduction URL

https://play.tailwindcss.com/7R6rwmsJQn

### Describe your issue

On Mac at leats tested in firefox developer edition and in chrome when the checkbox is disabled the border seems to be overlapped by the checkbox itself. see the blow image or the reproduction url
<img width="969" alt="Screenshot 2024-05-01 at 08 18 05" src="https://github.com/saadeghi/daisyui/assets/41805560/a48d06f5-443a-4705-a5a9-1b5a5a43b2ca">

PR #3010, opened 2024-05-01
Fix for #3009

tojaprice:
Added border-0 to the disabled state of the checkbox to fix the issue highlighted here: https://github.com/saadeghi/daisyui/issues/3009


saadeghi:
Thanks!

PR #3023, opened 2024-05-07
Supplement `.tab-active` modifier

seanpdoyle:
Supplement the `.tab-active` modifier to also support elements rendered with `[aria-selected="true"]`. When an element with `[role="tab"]` nested within a `[role="tablist"]` becomes "active" or "selected", its `[aria-selected]` attribute must be set to `"true"`. Similarly, when a tab is inactive, its `[aria-selected]` must be removed or set to `"false"` (according to the [WAI ARIA Guidelines][]).

Since all compliant tab implementations must set this attribute, and *not all* compliant tab implementations provide hooks to configure CSS Class List changes to be made alongside those attribute changes, this commit modifies the `tab.css` files to target either `.tab-active` or `[aria-selected="true"]` using the CSS [:is][] psuedo-class.

[WAI ARIA Guidelines]: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/
[:is]: https://developer.mozilla.org/en-US/docs/Web/CSS/:is

### DIFF
--- a/src/components/styled/tab.css
+++ b/src/components/styled/tab.css
@@ -10,7 +10,7 @@
   color: var(--tab-color);
   padding-inline-start: var(--tab-padding, 1rem);
   padding-inline-end: var(--tab-padding, 1rem);
-  &.tab-active:not(.tab-disabled):not([disabled]),
+  &:is(.tab-active, [aria-selected="true"]):not(.tab-disabled):not([disabled]),
   &:is(input:checked) {
     @apply border-base-content border-opacity-100 text-opacity-100;
   }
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks 👍  
Can you please undo the edit on the changelog file? It will be updated automatically after merge.

seanpdoyle:
@saadeghi done!

I also want to share that I was unable to get the `yarn playground` or `yarn dev` running locally, so the CSS changes were made without verifying their outcome locally. 

Is there somewhere I could learn more about local development? If not, could you verify they behave as you'd expect?

saadeghi:
Thanks. Yes I will test it before merge.  

For running on local, this guide should help: https://github.com/saadeghi/daisyui/blob/master/.github/CONTRIBUTING.md#building-on-local

PR #3061, opened 2024-06-01
Instruction to applying dark selector

lukababu:
I am just providing the instructions as read in #640 

I just wanted to be able to apply dark: selector to increase the border in light mode, and #640 was the solution which I think others would be looking for


saadeghi:
Thanks!

Issue #3060, opened 2024-05-30
bug: stepper (active step) z-index is higher then of the drawer


shahraship:
### What version of daisyUI are you using?

v4.11.1

### Which browsers are you seeing the problem on?

All browsers

### Reproduction URL

https://play.tailwindcss.com/DKbzSUjlXu

### Describe your issue

Active step should not have higher z-index then the drawer. It should show behind the drawer.

JohanAanesen:
@saadeghi I have solved this by swapping the z axis and using the negative z instead on the stepper, which leave the positive z-axis still available for the rest of the content. 
<img width="244" alt="image" src="https://github.com/user-attachments/assets/0c0c55df-f62d-42d5-9c53-2f8a758eb66e">
Might be a "cleaner" solution to the issue, but doesn't get rid of the z-axis entirely

PR #3070, opened 2024-06-05
#3060 - update drawer.css z-index

tojaprice:
Related to https://github.com/saadeghi/daisyui/issues/3060

Should prevent any elements in DaisyUI from having a higher z-index than the drawer.

### DIFF
--- a/src/components/styled/drawer.css
+++ b/src/components/styled/drawer.css
@@ -1,6 +1,7 @@
 .drawer {
   width: 100%;
   &-side {
+    @apply z-10;
     & > .drawer-overlay {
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks for the PR.

It is a quick fix but I think `10` is such a random value for z-index. This will potentially break the websites that are relying on the current behavior. We try to be neutral about the z-index values since it is something developers should decide about based on their content. 

Sometimes, based on the website content, we may need some components to appear on top of drawer and sometimes under the drawer for example we may want a toast notification to be visible on top of the drawer.

To fix the z-index issue of #3060 I prefer to find a solution to remove the z-index value from steps instead of increasing the z-index of drawer.

PR #3093, opened 2024-06-20
Update +page.svelte.md

garysassano:
Removed extra unneeded code.


saadeghi:
Extra unneeded? 🤔

garysassano:
Maybe I misunderstood then. I didn't need to use that extra code, and I saw the other components didn't have it, so I assumed it was there by mistake.

saadeghi:
Use the first example then:  https://daisyui.com/components/range/

The second example has these measurement lines, for people who want them. It's not extra:  
Screenshot 2024-06-22 at 8 59 45 PM

PR #3163, opened 2024-08-10
Added vertical ranges

### LINKED ISSUE #1483: feat: Vertical Range Slider
It'd be great if we had a standardized vertical range slider. I know this is rather tricky to do, but aren't framework there for this. :)

Here's a cross browser approach which should also be helpful when ignoring the JS wrap:
https://dev.to/konnorrogers/creating-a-vertical-slider-using-input-typerange-1pen 

What do you think?

TheBrenny:
Fixes #1483.

Adds styles for vertical Range sliders: `.range-vertical` and `.range-vertical.range-invert`.

I chose `.range-invert` because I'm not sure if there'll be an intent to add such a feature for all other sliders...

### DIFF
--- a/src/components/styled/range.css
+++ b/src/components/styled/range.css
@@ -95,18 +95,42 @@
     &.range-xs {
       width: 1rem;
       height: 100%;
+      &::-webkit-slider-runnable-track {
+        width: 0.25rem
+      }
+      &::-moz-range-track {
+        width: 0.25rem
+      }
     }
     &.range-sm {
       width: 1.25rem;
       height: 100%;
+      &::-webkit-slider-runnable-track {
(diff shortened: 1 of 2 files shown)


TheBrenny:
Well this is awkward... Apparently it worked and now it doesn't?

pdanpdan:
Check this one too: https://play.tailwindcss.com/PcCY6WIefZ
It only transform rotates the normal range slider, so you need to adjust the final width of the container to put multiple sliders side-by-side, but is uses very little styles

saadeghi:
Thanks for the PR.  
closed in favor of https://github.com/saadeghi/daisyui/pull/4554  
It will be in v5.6

PR #3164, opened 2024-08-10
Windows compatibility

TheBrenny:
Adds a very small amount of Windows compatibility to the build script.

This now Works on My Machine™️, but hopefully it hasn't broken how the script functions on other non-Windows hosts.

### DIFF
--- a/src/build.js
+++ b/src/build.js
@@ -1,4 +1,5 @@
-const { exec } = require("node:child_process")
+const { exec } = require("node:child_process");
+const { normalize } = require("node:path");
 const isWin = process.platform === "win32";
 const ext = isWin ? ".cmd" : "";
 const cat = isWin ? "type" : "cat";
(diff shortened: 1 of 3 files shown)


saadeghi:
Thanks for the PR!  
Isn't it better to use PowerShell or something similar witch supports Unix commands?

TheBrenny:
Powershell still isn't the default "shell" on windows. Running commands from Node like we're doing at the moment sort of opens up `cmd /C ...` (forgive me for not being perfectly technical). To use Powershell, we'd need to wrap all the commands in `pwsh -c ...` which is also just as feasible (I don't know why I didn't think of that to begin with!).

saadeghi:
PowerShell is pre installed on Windows.  
The "default" cmd.exe is not recommended for developers anyway. Just like "Notepad" the default text editor is not a good choice for writing code 😅
I mean... Isn't it better to use the right tool instead of adding OS specific `if` conditions to all the codebases on GitHub which are using Unix commands?

TheBrenny:
Hahaha, fair enough! I'll try and whip something up real quick!

PR #3165, opened 2024-08-10
Windows compatibility 2

TheBrenny:
Following from #3164, this is an update to the build script which now uses PowerShell. 

There's also some minor refactoring done for some of the `runCommand` calls to remove the erroneous `{stdio: []}` objects.

[Line 51](https://github.com/saadeghi/daisyui/compare/master...TheBrenny:daisyui:installer-fix#diff-abc8f6a58535fbae6adcdf1f8348cb6bed037902afdf5731f0070dd0cdb28753R47-R48) had to be broken up into two because Powershell kept reading the curly braces as a script block.

The changes to `package.json` make the switching between dev and prod platform agnostic by using NodeJS instead.

My builds also weren't working until I updated the vite config (windows read `/src/components/` as `C:\src\components\`), hence the change there.

Thanks for the heads up on using PowerShell! My world has low-key been opened up!

### DIFF
--- a/src/build.js
+++ b/src/build.js
@@ -1,59 +1,53 @@
 const { exec } = require("node:child_process");
-const { normalize } = require("node:path");
-const isWin = process.platform === "win32";
-const ext = isWin ? ".cmd" : "";
-const cat = isWin ? "type" : "cat";
 
 function runCommand(command) {
-    if(isWin) {
-        command = command.split(" ");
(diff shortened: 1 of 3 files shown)


saadeghi:
Thanks for the PR.  
The build steps and project structure is changed in v5. This workaround won't be needed anymore 💚

PR #3179, opened 2024-08-21
Update de.json

StaticFX:
Some words were translated to literal. The word "Store" in german, was translated as "to save". And the word "Theme" can be translated to the word "Thema" but, the word "Design" fits better in this context.


StaticFX:
That is true about design. Something like "Farbpalette" which would translate to colorpalette could also be a fitting word. Thema is more often used as "subject" instead of the actual theme of something.

saadeghi:
Let's use `Farbpalette` then.

StaticFX:
I have adjusted it in the correct places in my opinion

saadeghi:
Thanks @StaticFX

PR #3190, opened 2024-08-29
[docs] Add dark mode framework logos

lorenzolewis:
Followup to https://github.com/saadeghi/daisyui-images/pull/1#issuecomment-2318558691

- Change the shape of `exampleRepos` to hold images in an object so they can be referenced by key
- Add (TEST) dark Astro logo
- Refactor the section in `docs/install` that renders the frameworks to conditionally show a `primaryDark` image
- Add `darkMode` (and respective function) to the doc site's tailwind config to dynamically pull in and construct all dark themes

If this approach looks okay to you then let me know and I'll work on making a PR in https://github.com/saadeghi/daisyui-images to add the Astro dark image then update here.

### DIFF
--- a/src/docs/src/routes/(redirects)/tailwindplay/+page.server.js
+++ b/src/docs/src/routes/(redirects)/tailwindplay/+page.server.js
@@ -1,5 +1,5 @@
 export const ssr = false
 import { redirect } from "@sveltejs/kit"
 export function load() {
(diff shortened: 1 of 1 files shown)


lorenzolewis:
I also noticed that framework logos are also being used in a few other places such as the homepage so I can take a look at separate PRs for those using the same strategy as well.

PR #3192, opened 2024-08-31
Add tailwindcss as peer dep

rtritto:


saadeghi:
Why

rtritto:
_DaisyUI_ can't be used without _TailwindCSS_

saadeghi:
It can be used without Tailwind CSS.  
daisyUI package includes raw CSS files which people can directly use in their project.
It can also be used with other utility first CSS frameworks.

PR #3266, opened 2024-11-08
Dark mode selector update

RahulDey12:
it's will not work with class we should use `selector`

ref: https://tailwindcss.com/docs/dark-mode#customizing-the-selector


saadeghi:
Thanks.

>The selector strategy replaced the class strategy in Tailwind CSS v3.4.1.

https://tailwindcss.com/docs/dark-mode

Such a silent braking change 😶

Issue #3268, opened 2024-11-09
bug: accordion doesn't open/close on IOS using unstyled mode


c-h-o-n:
### What version of daisyUI are you using?

v4.12.14

### Which browsers are you seeing the problem on?

Safari iOS, Other

### Reproduction URL

https://stackblitz.com/edit/daisyui-react-vite-qhww2b?file=tailwind.config.js

### Describe your issue

I'm using daisyui with `styled: false` config value. The accordions on Safari IOS and Chrome IOS browsers (and on MacOS before Safari 15) don't open/close on click. It works if `styled` is set to `true` in the config.

The issue seems like the missing `width: 100%` property on the radio inputs. On IOS browsers the radio input fields won't take full width unless it's specified.

I didn't test this on android.

PR #3293, opened 2024-11-24
fix: #3268 unstyled collapse input size

c-h-o-n:
Closes: https://github.com/saadeghi/daisyui/issues/3268

Specifying width and height for collapse input to make sure it is taking up all space in grid using any browsers.


saadeghi:
Thank you!

PR #3324, opened 2024-12-17
readMe contributors and premium sponsors not showing issue fixed but 

faizanalibaig:
readMe contributors and premium sponsors not showing issue fixed but the size is not the same as before. The images are now showing. Please check the size issue now. : )

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,17 +25,18 @@
 
 ## Community
 
+
 <div align="center">
 
 Premium Sponsors  
-[![][sponsors]][opencollective]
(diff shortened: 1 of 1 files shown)


saadeghi:
This is a temporary cache issue from contrib.rocks.  
Instead of changing all the URLs, can you just add a `&` at the end of line 67?

Thanks

faizanalibaig:
Thank you for getting back to me. I haven’t worked with .md files this
extensively before. When I found this issue, I understood that it was
temporarily not working, but it was a great experience trying to resolve it
through an alternative approach.

saadeghi:
Fixed. thanks

Issue #3318, opened 2024-12-14
bug: Timeline Spacing


erikdemarco:
### What version of daisyUI are you using?

4.12.19

### Which browsers are you seeing the problem on?

All browsers

### Reproduction URL

https://daisyui.com/components/timeline/#timeline-with-icon-snapped-to-the-start

### Describe your issue

Just see this url in mobile, and you will see the spacing between timeline is not same:
https://daisyui.com/components/timeline/#timeline-with-icon-snapped-to-the-start

PR #3331, opened 2024-12-21
bug: Timeline Spacing #3318

kcherki:
Fix timeline style:
- Added a condition for the "md" size to the `<div class="$$timeline-end">` class to manage spacing on "md" screen sizes: replaced `mb-10` with `md:mb-10`.
- After checking the style, no conflicts were found; it was simply an adjustment for medium screen sizes (md).

**Otherwise, if there are any comments or if this is not the right place, please let me know with your feedback.**


src/components/unstyled/timeline.css
```diff
@@ -31,5 +31,5 @@
   @apply col-start-2 row-start-2;
 }
 .timeline-end {
-  @apply col-start-1 col-end-4 row-start-3 row-end-4 m-1 self-start justify-self-center;
+  @apply col-start-1 col-end-4 row-start-1 row-end-2 m-1 self-start justify-self-center;
```
saadeghi, line 34: Why?
This is breaking `timeline-end`
Before:
Screenshot 2024-12-21 at 11 13 15 PM

After:
Screenshot 2024-12-21 at 11 13 22 PM

src/docs/package-lock.json
saadeghi, line 1: Don't commit lockfiles or formatting please

saadeghi (changes requested):

src/components/unstyled/timeline.css
kcherki: Sorry for the confusion, this change was made only for testing and wasn’t meant to be sent. Thanks for pointing it out!

saadeghi: Can you explain what's the issue exactly? I can help.

PR #3334, opened 2024-12-24
Fix ESM export in documentation

max-programming:
A minor change because ESM has `export default` instead of `module.exports`


saadeghi:
Thanks!

PR #3345, opened 2024-12-26
Update diff.css

Mrahmani71:
The `Diff` component was not work working on RTL sites. While there may be alternative solutions, this fix offers a simple and effective solution.

### Changes:
- Applied a `direction: ltr` style to the `Diff` component to ensure proper alignment and functionality in RTL environments.

### Before (without `direction: ltr`):
image

### After (with `direction: ltr`):
image


saadeghi:
Thanks!

PR #3411, opened 2025-01-24
docs: add Vike install

rtritto:
Add Vike install to documentation of daisyUI v5.

Logo source: https://github.com/vikejs/vike-assets/blob/aef0b42307f9a0492355d8efd2933eb97bfddc6e/logo/favicon/vike-favicon.svg

### TODO
- [x] add create new Vike project command


rtritto:
@brillout Is there a command to create a new Vike project?

brillout:
Yes:

```bash
# Scaffold a Vike + vike-react app
pnpm create bati --react
# Scaffold a Vike + vike-vue app
pnpm create bati --vue
# Scaffold a Vike + vike-solid app
pnpm create bati --solid
```

magne4000:
> Yes:
> 
> ```shell
> # Scaffold a Vike + vike-react app
> pnpm create bati --react
> # Scaffold a Vike + vike-vue app
> pnpm create bati --vue
> # Scaffold a Vike + vike-solid app
> pnpm create bati --solid
> ```

You can directly add `--daisyui` in there too :)

rtritto:
@brillout @magne4000 I updated the doc, is there any change/review?

@saadeghi is there a generic Component like https://github.com/saadeghi/daisyui/blob/master/src/docs/src/components/homepage/Install.svelte? The purpose is to have a tab for each package manager.

brillout:
@rtritto Neat :100:. Idea: how about we break down the page into exactly two sections `## Scaffold new Vike app with daisyUI` and `## Add daisyUI to existing Vike app`?

rtritto:
> @rtritto Neat 💯. Idea: how about we break down the page into exactly two sections `## Scaffold new Vike app with daisyUI` and `## Add daisyUI to existing Vike app`?

Yes, this is another point. @saadeghi do you have a solution?

saadeghi:
> @saadeghi is there a generic Component like https://github.com/saadeghi/daisyui/blob/80d7a76bb6d4b09e86c3039a201eaf02c2e103a4/src/docs/src/components/homepage/Install.svelte? The purpose is to have a tab for each package manager.

Not yet, but if I add it in the future I will update the tutorials accordingly. Let's just use the default `npm` for now.

About the structure, can we follow the same structure as [other tutorials](https://v5.daisyui.com/docs/install/vite/)?
- Create a new Vike project _(no tailwind, no daisyUI, no bati, just `npm create vike`)_
- Install Tailwind CSS and daisyUI
  - Add Tailwind CSS to Vite config
  - Put Tailwind and daisyUI to your CSS file

This way people who already have a Vike project can simply skip first step. no need for option 1, option 2, etc.

rtritto:
> Create a new Vike project (no tailwind, no daisyUI, no bati, just `npm create vike`)

Bati is the tool, only for Vike, to generate a new Vike project.
It can use some options to dynamically add dependencies and code.
You can create a new Vike project with TailwindCSS and daisyUI and code with a single command (`npm create bati --- --react --daisyui`).
Without a Vike project, there is any sense to manually add TailwindCSS and daisyUI and code if the Bati can do that.
Imo, in the doc, would be better to have all options because the user can have different needs.

Anyway, I updated the doc removing the first option. Let me know to do a revert (with 2 options) or some change/review.

saadeghi:
I understand, I really like Bati as a tool and I use it myself.  
But the objective of this tutorial is not introducing new scaffolding tools.  
It's not a good idea for this tutorial page to have a special format or special tooling so let's keep it as similar as possible to other tutorial pages.

packages/docs/src/routes/(routes)/docs/install/vike/+page.md
```diff
@@ -0,0 +1,50 @@
+---
+title: Install daisyUI for Vike
+desc: How to install Tailwind CSS and daisyUI in a Vike project
+---
+
+### 1. Create a new Vike project
+
+Create a new Vike project in the current directory
+
+```:Terminal
```
saadeghi, line 10: Please use `npm create vike`. We don't want to introduce a new scaffolding tool because then we have to explain what is which is not the objective of this page.
Also, no need for flags for different frameworks.

saadeghi (changes requested):

packages/docs/src/routes/(routes)/docs/install/vike/+page.md
rtritto: Done

saadeghi:
Thank you ❤️

Issue #3437, opened 2025-01-30
docs: Daisy UI Buttons with any HTML tags


saent-x:
### On which page do you see this issue?

https://v5.daisyui.com/components/button/

### Describe the issue

The example in the v5.0.0-beta.5 is not working as is;

```
...
<input type="radio" aria-label="Radio" class="btn" />
<input type="checkbox" aria-label="Checkbox" class="btn" />
<input type="reset" value="Reset" class="btn" />
```

the tag "<input type="reset" value="Reset" class="btn" />" does not clear when clicked!!

It only works when used with a form, it should be added to the example;
```
<form style="display:contents">
     <input type="radio" aria-label="Radio" class="btn"> 
     <input type="checkbox" aria-label="Checkbox" class="btn"> <input type="reset" value="Reset" class="btn">
</form>
```

### What browsers are you seeing the problem on?

Chrome

saadeghi:
The example is about how the element looks like using the `btn` class.  

The form tag is removed from the code example intentionally to make it clear form tag is not required for the button to look the way it looks.  

It's up to you where and how to use them.

Let me know if you have any questions

PR #3438, opened 2025-01-30
Saent x patch 1

saent-x:
The previous content just shows;
```
<input type="radio" aria-label="Radio" class="$$btn" />
<input type="checkbox" aria-label="Checkbox" class="$$btn" />
<input type="reset" value="Reset" class="$$btn" />
```
without the form element and reset doesn't work without it, updated the docs to include the form element like below;
```
<form style="display:contents">
  <input type="radio" aria-label="Radio" class="btn" />
  <input type="checkbox" aria-label="Checkbox" class="btn" />
  <input type="reset" value="Reset" class="btn">
</form>
```


saadeghi:
Sorry I can't merge this. 

https://github.com/saadeghi/daisyui/issues/3437#issuecomment-2625168750

saent-x:
@saadeghi okay, understood

Issue #3430, opened 2025-01-29
bug: v5 beta3 CDN Styling issues


Mashiane:
### What version of daisyUI are you using?

v5

### Which browsers are you seeing the problem on?

Edge

### Reproduction URL

https://play.tailwindcss.com/ndb1oBR8jW

### Describe your issue

Hi there

Just tried the latest beta 3 and tailwind 4 via CDN

1. The collapse border and background have not been applied.

aglio:
Hi, @saadeghi!
I ran into the same issue, so I proposed a fix. I'd appreciate it if you could merge it if everything looks good!
https://github.com/saadeghi/daisyui/pull/3447

PR #3447, opened 2025-02-02
fix: #3430 - Remove @layer from the CDN build

aglio:
Fix: #3430 

## Problem Cause
https://developer.mozilla.org/en-US/docs/Web/CSS/@layer#description
> Styles that are not defined in a layer always override styles declared in named and anonymous layers.

### Issue
The styles generated by `https://cdn.tailwindcss.com` are not defined within a `@layer`, which gives them the highest priority. As a result, if all daisyUI styles are defined inside a `@layer`, Tailwind CSS's base styles will always take precedence.

_Development__Rails_Template

So, I propose removing `@layer` in the CDN build to prevent this issue.

### Causing commit
https://github.com/saadeghi/daisyui/commit/a25eae96172fa7243bad8640b7a0b73e19d03e66

## Issue Reproduction URL (Before Fix)
https://stackblitz.com/edit/daisyui-cdn-8edfkhem?file=index.html

### DIFF
--- a/packages/daisyui/build.js
+++ b/packages/daisyui/build.js
@@ -27,7 +27,7 @@ async function generateFiles() {
         states: ["hover"],
       }),
     !isDev && generateThemeFiles({ srcDir: "src/themes", distDir: "theme" }),
-    !isDev && generateRawStyles({ srcDir: "../src/base", distDir: "../base", layer: "base" }),
+    !isDev && generateRawStyles({ srcDir: "../src/base", distDir: "../base" }),
     !isDev &&
       generateRawStyles({
         srcDir: "../src/components",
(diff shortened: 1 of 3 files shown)


saadeghi:
Can you us the new Tailwind CSS cdn file?  
https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4

Docs https://v5.daisyui.com/docs/cdn/
Tailwind CSS CDN docs: https://tailwindcss.com/docs/installation/play-cdn

https://cdn.tailwindcss.com is the old CDN file. The new one uses layers.

aglio:
@saadeghi Oh, my bad, it looks like using https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4 fixed it! Thanks!
I should have checked the documentation properly… I'll go ahead and close this PR.

It looks like #3430 is the same issue.

PR #3450, opened 2025-02-03
Update dark variant

wiseguy1394:
@variant for dark is not supported by newest tailwindcss plugin.


saadeghi:
Looks like it is still supported and `@variant` get converted to `@custom-variant` internally.  
source: https://github.com/tailwindlabs/tailwindcss/pull/15663

But let's merge this anyway to be aligned with the new syntax. Thanks 👍

PR #3456, opened 2025-02-05
Added examples of using gradient backgrounds on buttons

Avaray:
I’m not sure if `<style>` will be processed correctly in the documentation page.
If you think this button example is unnecessary, feel free to reject the PR.

The buttons will look like this: https://play.tailwindcss.com/C7EgefbQww

### DIFF
--- a/packages/docs/src/routes/(routes)/components/button/+page.md
+++ b/packages/docs/src/routes/(routes)/components/button/+page.md
@@ -495,3 +495,33 @@ classnames:
 </button>
 
 ```
+
+### ~Button with gradient background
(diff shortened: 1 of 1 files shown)


saadeghi:
I wish there was a way to define keyframes using utility classes… so people could just copy the HTML, without needing to add style to CSS file.

Avaray:
Yep, there’s currently no way to do that with Tailwind CSS.

saadeghi:
Thanks for the PR but this is not ideal to add custom CSS into the CSS for this to work. It would be nice if it was possible with utility classes.

PR #3461, opened 2025-02-05
Corrected example code by removing incorrect colons

Avaray:
Removed unnecessary colons after class names in the example code, which were causing syntax errors.


saadeghi:
😅👍

PR #3469, opened 2025-02-06
Update custom theme documentation

hajamie:
The example now shows how to use the new custom theme (add it to the themes list in the daisyui plugin configuration) and enables it (default: true).

### DIFF
--- a/packages/docs/src/routes/(routes)/docs/themes/+page.md
+++ b/packages/docs/src/routes/(routes)/docs/themes/+page.md
@@ -102,10 +102,12 @@ To add a new theme, use `@plugin "daisyui/theme" {}` in your CSS file, with the 
 
 ```css:app.css
 @import "tailwindcss";
-@plugin "daisyui";
+@plugin "daisyui" {
+  themes: mytheme;
+}
 @plugin "daisyui/theme" {
   name: "mytheme";
-  default: false; /* set as default */
+  default: true; /* set as default */
   prefersdark: false; /* set as default dark mode (prefers-color-scheme:dark) */
   color-scheme: light; /* color of browser-provided UI */
 


packages/docs/src/routes/(routes)/docs/themes/+page.md
```diff
@@ -102,10 +102,12 @@ To add a new theme, use `@plugin "daisyui/theme" {}` in your CSS file, with the
 
 ```css:app.css
 @import "tailwindcss";
-@plugin "daisyui";
+@plugin "daisyui" {
+  themes: mytheme;
```
saadeghi, line 106: This doesn't do anything.

`@plugin "daisyui"{}` is for built-in themes.  
`@plugin "daisyui/theme" {}` is for custom themes.  

It's separated because Tailwind's plugin API doesn't allow multi nested properties as plugin options.  

Please only commit `default: true;`

saadeghi (changes requested):

packages/docs/src/routes/(routes)/docs/themes/+page.md
hajamie: I took out this bit in my globals.css and my custom theme didn't load.  
```diff
- @plugin "daisyui" {
-   themes: <ThemeName>;
- }
+ @plugin "daisyui";
```
Is it an issue somewhere else?

This is what I currently have:

```
@import "tailwindcss";
@plugin "daisyui" {
  themes: Name;
}
@plugin "daisyui/theme" {
  name: "Name";
  default: true;
  prefersdark: false;
  color-scheme: light;

  /* Base colors */
  --color-base-100: #ffffff;
  --color-base-200: #f0f0f0;
/* etc. */
}

@config '../tailwind.config.ts';

/* Then some things the tailwind upgrade added */
```

saadeghi: It does work: https://play.tailwindcss.com/3w61U3AdJa?file=css  
- I just added the [missing variables](https://v5.daisyui.com/docs/themes/#-2).

- [for Tailwind Play only] I added `@beta` in `daisyui@beta` because it's required for Tailwind Play until stable version comes.  

- If it's not working, can you share that `tailwind.config.ts` your importing?

Btw here
```
@plugin "daisyui" {
  themes: Name;
}
```
instead of `themes: Name;` use instead of `themes: false;`  
``
`themes` can be `false`, or `all`, or a list of built-in themes. 

docs: https://v5.daisyui.com/docs/config/#themes  
I just added more examples to make it more clear.

PR #3499, opened 2025-02-16
Adjusted default dock component (dock-md) sizes for A11Y compliance (min 12px) and ensured linear scaling with size class.

IdrisCelik:
These changes comply with A11Y best practices for the default dock size, which is dock-md. 
The dock sizes now scale correctly, instead of md being bigger than lg. While respecting the minimum (xs) and maximum (xl) sizes set by the maintainer.
All of this while ensuring it still looks good to the eye on different devices.

### DIFF
--- a/packages/daisyui/src/components/dock.css
+++ b/packages/daisyui/src/components/dock.css
@@ -29,7 +29,7 @@
     }
 
     .dock-label {
-      font-size: 0.6875rem;
+      font-size: 0.75rem;
     }
   }
 }
@@ -61,7 +61,7 @@
   }
 
   .dock-label {
-    font-size: 0.625rem;
+    font-size: 0.6875rem;
   }
 }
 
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks for the PR.  
However I will keep the minimum as 10px.  

11 is Apple's minimum:   
https://developer.apple.com/design/human-interface-guidelines/typography#Ensuring-legibility

10 is Google's minimum:  
https://m3.material.io/styles/typography/type-scale-tokens#2a57c8f0-d45b-470a-984f-eb4f89f425fa

The a11y best practices who suggest 12px are very very old. Back then displays were low quality and couldn't show small font sizes properly.  
Nowadays even the cheapest phone displays can show 10px text properly.  

Of course font family and color effects readability as well, but font size of 10 is a legit minimum.

PR #3503, opened 2025-02-17
V5: Fixed loading component default size

OmkarOza:
For the default size, the width is defined using the Tailwind CSS `w-6` class. However, for other sizes, including `md`, the width is specified using a CSS variable.

**Changes:**

 - Removed `w-6` (fixed width class).
- Added width: calc(var(--size-selector, 0.25rem) * 6); for dynamic sizing.

**Reason for Change:**
- Enables customization of the width based on `--size-selector`, improving responsiveness and adaptability.

Let me know if you’d like any refinements!


saadeghi:
Thanks!

PR #3534, opened 2025-02-22
re-add specificity to .list-row, only for .list-col-grow functionality

### LINKED ISSUE #3531: bug: List component has alternating columns grow with list-col-grow

5.0.0-beta.8

Firefox

### Reproduction URL

https://v5.daisyui.com/components/list/#list-third-column-grows

### Describe your issue

The `list-col-grow` modifier in the List component does not make a specific column inside a `list-row` grow, and seems to affect the grow modifier in subsequent `list-row` elements inside the same `list` too.

It seems that the `grow` modifier is applied to one column later in each subsequent element.

The effect is also visible in the documentation page of Lists (https://v5.daisyui.com/components/list/#list-third-column-grows):
Image

The example from the docs with grid visualization from Firefox development tools:
Image

I am using Firefox 135.0.1

Acccent:
fixes #3531

### DIFF
--- a/packages/daisyui/src/components/list.css
+++ b/packages/daisyui/src/components/list.css
@@ -6,6 +6,11 @@
     @apply rounded-box relative grid grid-flow-col gap-4 p-4;
     word-break: break-word;
     grid-template-columns: var(--list-grid-cols);
+    :not(.list-col-wrap) {
+      @apply row-start-1;
+    }
+  }
+  .list-row {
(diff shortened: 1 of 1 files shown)


saadeghi:
This doesn't fix the issue.

saadeghi:
Issue is fixed in Tailwind CSS `4.0.9`  
https://play.tailwindcss.com/eLpMfOVTzl

Thanks for the PR.

PR #3541, opened 2025-02-24
Update .prettierrc

DaviDevMod:
Put `"prettier-plugin-tailwindcss"` last, as per [docs](https://github.com/tailwindlabs/prettier-plugin-tailwindcss?tab=readme-ov-file#compatibility-with-other-prettier-plugins).


saadeghi:
Thanks!

PR #3550, opened 2025-02-25
Add "bun.lock" to .gitignore

DaviDevMod:
Bun v1.2 changed the default lockfile format to the text-based `bun.lock`.
[Source](https://bun.sh/docs/install/lockfile)


DaviDevMod:
It was supposed to be a single commit PR, I'm going to mess with my branch to remove the previous commit.

saadeghi:
Thanks!

PR #3552, opened 2025-02-25
Add .prettierignore

### LINKED ISSUE #3542: docs: Duplicate classes

https://v5.daisyui.com/components/collapse/

### Describe the issue

The issue may actually interest the whole repo, not just the docs.

[Here](https://github.com/saadeghi/daisyui/blob/8401840a94c2dba741c83bac9398ed500449435f/packages/docs/src/routes/(routes)/components/collapse/%2Bpage.md?plain=1#L103) you can find duplicate `border border-base-300` classes:
```md
<div tabindex="0" class="collapse border border-base-300 bg-base-100 border border-base-300 collapse-arrow">
```
There are more duplicates in the same file, and possibly in the rest of the repo.

While looking into the issue, I noticed this #3541 , but it didin't fix the problem.
I don't have time to spend on the problem, so I'm just gonna leave you with this ticket.

_No response_

DaviDevMod:
Fixes #3542

But also the prettier setup of the repo in general.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,6 @@
+# Package Managers
+package-lock.json
+pnpm-lock.yaml
+yarn.lock
+bun.lockb
+bun.lock


saadeghi:
Thanks!

PR #3553, opened 2025-02-26
Duplicate classes

DaviDevMod:
Complete fixing #3542


DaviDevMod:
There is a lot of formatting going on, I think it's legit, but you do you.

PR #3557, opened 2025-02-27
fix(getFileNames): Fix the recursive call to add the extension

swotex:
This fixes an issue where the recursive call to `getFileNames` mistakenly passed `recursive` instead of `extension`. 
Now, the function correctly uses the `extension` parameter in recursive calls.

This PR is targeted for the `v5` branch.


saadeghi:
Nice 🔥

PR #3599, opened 2025-03-05
fix(button.css): update btn-outline styles to include checked state

aforcada:
When a checkbox or radio is checked and styled as btn-outline with btn-[color], the text color does not display correctly.

Before
image

After
image


saadeghi:
Thanks. It will be in 5.0.1 soon

PR #3640, opened 2025-03-10
Docs: Fix disabled button HTML differing from preview.

whitetxt:
Removed the extra shown disabled button with btn-outline instead of adding it to the shown HTML, as btn-outline does not change how the disabled button looks.


saadeghi:
Thanks 👍

PR #3649, opened 2025-03-11
v5 (Docs): Updated WordPress installation guide

suabahasa:
Follow-up update of PR #3526

Now we can use the [esm.sh](https://esm.sh) as the default CDN provider


saadeghi:
Nice 👍

PR #3669, opened 2025-03-13
(feat): Add validator token to prevent success color

brunorodmoreira:
The purpose is to support a modifier token in the `validator` component that allows developers to opt-out the success color for valid inputs.

Behavior:
image

Solves: #3668

### DIFF
--- a/packages/daisyui/src/components/validator.css
+++ b/packages/daisyui/src/components/validator.css
@@ -1,12 +1,14 @@
 .validator {
-  &:user-valid,
-  &:has(:user-valid) {
-    &,
-    &:focus,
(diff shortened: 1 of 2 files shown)


williamhnyohei (approved):

saadeghi:
Thanks for the PR. Instead of adding new class names (like `validator-invalid`), it's now possible to use `validator` with Tailwind CSS variants like `user-valid:validator` or `user-invalid:validator` so you can get the green or red colors only for valid or invalid value (not both).

Example with daisyUI `5.0.34`: https://play.tailwindcss.com/8KzLCBB3YH

PR #3672, opened 2025-03-13
feat: #3671 add a CSS var to customize the hover color

### LINKED ISSUE #3671: feat: add a CSS var for the hover state in every Color.
Hi
I have a color theme of my own where the hover mechanism, which is a calculation of a new color out of the initial color, does not create a color that I like. Because of this I add a lot of hover:bg-ColorThatILike (linked to a new color CSS var that I create) in my code.

What I would find useful would be to already have some CSS var for the hover like a --color-primary-hover, a bit like there is a foreground color with the name --color-primary-content. This CSS var could override the bg color in hover state and we could keep the current behavior in normal cases. 
BR

ExFlo:
closes #3671 

It also change a bit the behavior of the CSS var that depends on the btn-color CSS Var as now in hover state those CSS var will depends on the hover btn-color value.


ExFlo:
I need to add tests.
We also need to discuss if it is needed, the impact on the focus-visible + hover combo...
This is a WIP of course. 
I think it improve the capability but it is more variables...
Let's maybe discuss this in the discussion I opened for it.
BR

saadeghi:
Thanks for the PR.  
I'm afraid I can't merge this unfortunately.  
I wished we talked about it before you putting the effort for PR...

Hover color would be only used for button and link. It's not efficient to register these color names to Tailwind just for two component.

Modifying hover color of butter would be same amount of effort if it's from a CSS variable or using a utility class. But the cost of adding all these colors is irreversible.

PR #3693, opened 2025-03-17
fix: #3692 provide an unminified bundle for the standalone plugin

Gazler:
This adds a new `build:standalone` which builds the plugin using rollup, which is the same tool used by jsDelivr.

The bundle is built to `packages/daisyui/daisyui.js` but has not been included in this PR to make the diff smaller.

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "scripts": {
     "build": "cd packages/daisyui && bun run build",
     "build:dev": "cd packages/daisyui && bun run build --dev",
+    "build:standalone": "bun run rollup -c",
     "build:docs": "bun run build && cd packages/docs && bun run build",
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks for the PR 💚  
Instead of Rollup, I used Bun, as it was already the build tool.  

Bundle files are now available on each release:  
https://github.com/saadeghi/daisyui/releases
```
curl --sLO https://github.com/saadeghi/daisyui/releases/latest/download/daisyui.js
```

Also updated [docs](https://daisyui.com/docs/install/standalone/) to use this file instead of esm.run

Gazler:
Thanks for implementing this! I like the bun version, I only used rollup because that's what esm.run was using.

PR #3733, opened 2025-03-29
fix(docs/seo): Remove VITE_ROOT prefix from SEO image paths

yug1224:
Fixed the image URL generation logic for `og:image` and `twitter:image` tags in the documentation site's SEO component.

Removed the call to `readEnv("VITE_ROOT")` because the value of `$props().img` is always passed as an absolute path URL.


saadeghi:
Thanks

PR #3761, opened 2025-04-07
fix: collapse-content transition duplicate, transition typo

mayerzx:
After the update to latest version, still one of the line has a typo that I changed with this PR + an additional duplicate key


saadeghi:
Thank you

PR #3763, opened 2025-04-08
Add `aria-invalid` to the validator error selectors

TheOnlyTails:
This change would allow users to set their own validator states not based on native HTML validation (which can be limited), while not adding any additional classes.
It also encourages using the existing `aria-invalid` property, increasing accessibility.


saadeghi:
Thanks!

PR #3777, opened 2025-04-12
"Spotiwhy" replaced with "Spotify"

developerbrm:
- "Spotiwhy" replaced with "Spotify"


saadeghi:
Thanks but it's not a typo

PR #3820, opened 2025-04-25
fix: button hover state persisting on mobile screen. closes: #3817

### LINKED ISSUE #3817: bug: Soft, outline and dash buttons remain in hover state after being touched on iOS

https://daisyui.com/components/button/

5.0.28

Safari iOS

### Describe your issue

Open DaisyUI website on iOS, go to components -> button
Scroll down to soft buttons click one => it remains in hover state after click
Click outside of the button => it returns to the normal state. 

This seems to also repro on Chrome when enabling the mobile view (cmd + shift + m) from the dev tools.

sulimanbenhalim:
## Problem
Fixes #3817 

On mobile devices (iOS Safari and others), when you tap soft, outline, or dash buttons, they get stuck in hover state. The button stays colored until you tap somewhere else on the screen.

## Fix
I added CSS that targets touch devices (using `hover: none` media query) to fix the three button types:
- soft buttons
- outline buttons
- dash buttons

Now the buttons go back to normal right after you tap them.

## Testing
I tested this on iPhone with Safari and with Chrome's mobile screen mode.

### DIFF
--- a/packages/daisyui/src/components/button.css
+++ b/packages/daisyui/src/components/button.css
@@ -187,6 +187,24 @@
     --btn-border: var(--btn-color);
     --btn-noise: none;
   }
+  
+  @media (hover: none) {
+    &:hover:not(
+        .btn-active,
+        :active,
+        :focus-visible,
+        :disabled,
+        [disabled],
+        .btn-disabled,
+        :checked
+      ) {
+      --btn-shadow: "";
+      --btn-bg: #0000;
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks. I'm trying to find a way to reduce the repetition if possible…

saadeghi:
> I wonder if we can find a way to reduce the repetition

Probably not…

Thanks for the PR!

PR #3821, opened 2025-04-25
fix: tooltip showing when data-tip or tooltip-content is empty. closes: #3819

### LINKED ISSUE #3819: bug: tooltip shows a part of tip with blank or missing data-tip/tooltip-content

https://play.tailwindcss.com/28GLwWiciC

v5.0.27

All browsers

### Describe your issue

If no message is provided in `data-tip` or `data-content`, the tooltip shows an artifact on hover. In v4, nothing was displayed in this case

Image

sulimanbenhalim:
## Problem
Fixes #3819

When a tooltip has an empty data-tip attribute or empty tooltip-content, it still shows a small artifact on hover. This looks like a UI bug.

## Fix
I added CSS rules to hide the tooltip when:
- data-tip attribute is missing
- data-tip attribute is empty
- tooltip-content is empty
- tooltip-content only has empty children

Now empty tooltips don't show up at all, which is what users would expect.

## Testing
I tested this with examples from the issue:
- Tooltip with empty data-tip=""
- Tooltip with empty tooltip-content

Before: Empty tooltips showed a strange artifact
After: Empty tooltips stay completely hidden

### DIFF
--- a/packages/daisyui/src/components/tooltip.css
+++ b/packages/daisyui/src/components/tooltip.css
@@ -74,8 +74,8 @@
 
 .tooltip {
   &.tooltip-open,
-  &[data-tip]:hover,
-  &:hover,
+  &[data-tip]:not([data-tip=""]):hover,
+  &:not(:has(.tooltip-content:empty)):has(.tooltip-content):hover,
   &:has(:focus-visible) {
     > .tooltip-content,
     &[data-tip]:before,


sulimanbenhalim:
Hey @saadeghi I updated the code using nesting as requested 

this covers:

- Elements with no data-tip attribute
- Elements with empty data-tip=""
- Elements without tooltip-content div

Note: This does not cover cases with empty tooltip-content divs like `<div class="tooltip-content"></div>`.
That's why the initial solution covered all 4 cases

sulimanbenhalim:
> That's ok. Since developers can decide when to render that div. 

You are right I think I just overthought the solution by focusing too much on special cases.

> Also, `:empty` selector may not work all the time, especially when HTML is being rendered by frameworks and may include spaces inside the div, in that case it won't be counted as `:empty`

I tested with whitespace in the `div` and `:empty` still catches it correctly, so now it is working as intended you can double check, peace!

PR #3829, opened 2025-04-30
add `touch-action: manipulation` to button.css

81reap:
- disable double-tap to zoom on buttons
- speed up click/tap events on buttons

# Links
- https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action#manipulation


81reap:
`touch-action: none` would also achieve the same results, but `manipulation` felt more appropriate. LMK if there are any other components this would be good to add to :)

saadeghi:
Thanks!

Issue #3843, opened 2025-05-04
bug: tab-content overflows when tabs has height


kuboon:
### Reproduction URL (Required)

https://play.tailwindcss.com/dAcF6M87N2?file=css

### What version of daisyUI are you using?

v.5.0.35

### Which browsers are you seeing the problem on?

Firefox, Safari

### Describe your issue

see playground.
https://github.com/saadeghi/daisyui/blob/v5.0.35/packages/daisyui/src/components/tab.css#L45
remove `h-full` and add `height: calc(100% - var(--size-field, 0.25rem) * 10);` will fix this.

I think it would be better to define var for tab height.

kuboon:
I'm ready to create PR for this issue https://github.com/kuboon/daisyui/pull/1

saadeghi:
That would be great. thank you

PR #3844, opened 2025-05-04
fixed bug: tab-content overflows when tabs has height #3843

rounakkraaj-1744:
fixed the issue addressed by @kuboon....made the respective changes in the codebase....

eagerly waiting for the pull request to be merged!!!

### DIFF
--- a/packages/daisyui/src/components/tab.css
+++ b/packages/daisyui/src/components/tab.css
@@ -41,10 +41,11 @@
   &:checked,
   &:is(label:has(:checked)),
   &:is(.tab-active, [aria-selected="true"]) {
-    & + .tab-content {
-      @apply block h-full;
-    }
+  & ~ .tab-content {
+    @apply block;
+    height: calc(100% - var(--tab-height, 2.5rem));
   }
+}
   &:not(:checked, label:has(:checked), :hover, .tab-active, [aria-selected="true"]) {
     @apply text-base-content/50;
   }


sulimanbenhalim (changes requested):
I see that you're using `height: calc(100% - var(--tab-height, 2.5rem));` with a fixed height value. Here's an alternative approach that leverages daisyUI's existing variable system:

Instead of creating a new `--tab-height` variable with a hardcoded fallback, we could use the existing `--size-field` variable which already accounts for different tab sizes:

```css
&:is(.tab-active, [aria-selected="true"]) {
  & + .tab-content {
    @apply block;
    height: var(--tab-content-height, 100%);
  }
}

.tab-content {
  --tab-content-height: calc(100% - var(--size-field, 0.25rem) * 10);
  /* ... existing properties ... */
}

/* Size-specific overrides */
.tabs-xs .tab-content {
  --tab-content-height: calc(100% - var(--size-field, 0.25rem) * 6);
}

.tabs-sm .tab-content {
  --tab-content-height: calc(100% - var(--size-field, 0.25rem) * 8);
}

.tabs-lg .tab-content {
  --tab-content-height: calc(100% - var(--size-field, 0.25rem) * 12);
}

.tabs-xl .tab-content {
  --tab-content-height: calc(100% - var(--size-field, 0.25rem) * 14);
}

kuboon:
@sulimanbenhalim how about do like this?
```css
.tabs-xs {
  :where(.tab) {
    --tab-height: calc(var(--size-field, 0.25rem) * 6);
    height: var(--tab-height);
  }
}
/* do same to other sizes */

&:is(.tab-active, [aria-selected="true"]) {
  & + .tab-content {
    @apply block;
    height: calc(100% - var(--tab-height));
  }
}

sulimanbenhalim:
@kuboon I see your approach - it's clever! Both solutions fix the issue, but yours has a nice advantage in that it's more DRY

sulimanbenhalim (changes requested):
you need to define the `--tab-height` variable for each tab size variant. Currently you're using a fallback value of 2.5rem which may not match actual tab heights

kuboon:
Thank you @rounakkraaj-1744 @sulimanbenhalim 

I've assigned my issue #3843  by @saadeghi  and made another PR #3873

PR #3856, opened 2025-05-08
fix: scope [data-theme] selectors to root option. closes: #3854

### LINKED ISSUE #3854: bug: `root` option leaks colors to non-root selector, with data-theme set

https://play.tailwindcss.com/bWgQcMcaNB

latest (5.0.35)

All browsers

### Describe your issue

When using a specific `root` option, I noticed the selector doesn't scope `data-theme`, which is used/recommended in other places (i.e. https://github.com/abereghici/remix-themes/blob/904f5db05bef1b9110fdc8492aff7db3df556a37/README.md?plain=1#L97), so embedding a daisyUI theme with potentially different colors can override existing styles.

So if an element has:

```html
<div data-theme="light">
  <button class="btn btn-primary">Primary</button>
</div>
```

it will get the styles applied despite not being under a `#root` ancestor

sulimanbenhalim:
## Problem
Fixes #3854

When using a custom root option, elements with data-theme attributes outside the root still receive theme styles. This causes unexpected styling when embedding daisyUI themes in different parts of an application.

## Fix
Changed the selector construction to properly scope data-theme selectors to the specified root element, preventing theme styles from leaking outside the root.

## Testing
Tested with the reproduction case from issue #3854 in multiple browsers.

### DIFF
--- a/packages/daisyui/functions/pluginOptionsHandler.js
+++ b/packages/daisyui/functions/pluginOptionsHandler.js
@@ -22,7 +22,7 @@ export const pluginOptionsHandler = (() => {
     const applyTheme = (themeName, flags) => {
       const theme = themesObject[themeName]
       if (theme) {
-        let selector = `${root}:has(input.theme-controller[value=${themeName}]:checked),[data-theme=${themeName}]`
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks for the PR.  
Explained here https://github.com/saadeghi/daisyui/issues/3854

PR #3873, opened 2025-05-12
fix: tab-content overflows when tabs has height. closes: #3843

### LINKED ISSUE #3843: bug: tab-content overflows when tabs has height

kuboon:
closes #3843

### DIFF
--- a/packages/daisyui/src/components/tab.css
+++ b/packages/daisyui/src/components/tab.css
@@ -17,9 +17,10 @@
   --tab-radius-ee: 0;
   --tab-order: 0;
   --tab-radius-min: calc(0.75rem - var(--border));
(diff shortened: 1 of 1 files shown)


kuboon:
@saadeghi thank you, I've fixed the problem.

PR #3883, opened 2025-05-15
fix: disabled button border inconsistency in join component. closes: #3877

### LINKED ISSUE #3877: bug: disabled button should not change border size

https://play.tailwindcss.com/mziC3eOkBa

v5

_No response_

### Describe your issue

Image

sulimanbenhalim:
## Problem
Fixes #3877

Disabled buttons in join components don't maintain consistent border styling because they're excluded from receiving the negative margins applied to regular buttons.

## Fix
Removed the disabled button exclusions (`:disabled`, `[disabled]`, `.btn-disabled`) from the CSS selector in join.css so all buttons receive the same negative margin treatment.

## Testing
Tested with various join component configurations including both enabled and disabled buttons.

### DIFF
--- a/packages/daisyui/src/utilities/join.css
+++ b/packages/daisyui/src/utilities/join.css
@@ -68,7 +68,7 @@
 }
 
 .join-item {
-  &:where(*:not(:first-child, :disabled, [disabled], .btn-disabled)) {
+  &:where(*:not(:first-child)) {
     margin-inline-start: calc(var(--border, 1px) * -1);
     margin-block-start: 0;
   }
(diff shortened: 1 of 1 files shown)


saadeghi (changes requested):
Thanks for the PR but this causes another issue:  
Before: https://play.tailwindcss.com/LlDimNd8cE?file=css  
After: https://play.tailwindcss.com/GlYDarHz6q?file=css

sulimanbenhalim:
> Thanks for the PR but this causes another issue: Before: https://play.tailwindcss.com/LlDimNd8cE?file=css After: https://play.tailwindcss.com/GlYDarHz6q?file=css

I have changed the implementation check it out now, I can't think of any other way to solve this

image

PR #3887, opened 2025-05-16
fix: URL and email inputs maintain LTR direction in RTL context. closes: #3886

### LINKED ISSUE #3886: bug: URL and email inputs with icons don't maintain LTR in RTL mode

https://daisyui.com/components/input/?lang=ar#url-with-icon-and-validator   https://daisyui.com/components/input/?lang=ar#email-input-with-icon-and-validator

5.0.35

All browsers

### Describe your issue

In RTL mode, URL and email inputs with icons reverse alignment, which is incorrect since these inputs should always display LTR content. URLs and emails are universally written with Latin characters and should maintain LTR properties even in RTL contexts.

Image

Image

I can fix this by adding CSS rules to ensure URL and email inputs maintain consistent LTR behavior in all languages.

You can assign this to me

sulimanbenhalim:
## Problem
Fixes #3886

URL and email inputs with icons don't maintain their LTR direction when used in RTL contexts. Since URLs and email addresses are universally written with Latin characters, they should always be displayed left-to-right, regardless of the page's text direction.

## Fix
Added a CSS rule that explicitly sets `direction: ltr;` for URL and email input types.

## Testing
Tested with URL and email inputs with icons on the daisyUI demo site with Arabic language (RTL) enabled, confirming correct left-to-right text direction and proper icon placement.

### DIFF
--- a/packages/daisyui/src/components/input.css
+++ b/packages/daisyui/src/components/input.css
@@ -28,6 +28,11 @@
     }
   }
 
+  :where(input[type="url"]),
+  :where(input[type="email"]) {
+    direction: ltr;
+  }
+
   :where(input[type="date"]) {
     @apply inline-block;
   }


saadeghi:
Thanks!

PR #3894, opened 2025-05-18
fix(collapse): min-height 1lh if supported

### LINKED ISSUE #3885: bug: The collapse input/title has a fixed min-height applied with high specificity

https://play.tailwindcss.com/30bBExaX98

v5.0.35

All browsers

### Describe your issue

The height of the collapse input and title is fixed at `3.75rem`, making it difficult to override from the outside using a simple `min-h-0` class.

Image

Related: #191

I don't know exactly why the value is set to `3.75rem`; the documentation doesn't explain it. However, it would be more practical to use `1lh`, considering that TailwindCSS v4 targets modern browsers by default, making this a viable unit. Still, we can apply a fallback value as well.

* [Line height units](https://developer.mozilla.org/en-US/docs/Learn_web_development/Core/Styling_basics/Values_and_units#line_height_units)
* [Can I use `lh` unit?](https://caniuse.com/mdn-css_types_length_lh)

https://play.tailwindcss.com/vozh54caSy?file=css

rozsazoltan:
Fixes #3885 

I believe the `3.75rem` min-height was set for safety reasons. However, this makes it impractical to use smaller font sizes, such as the default `text-xs`, and requires strong overrides.

The `1lh` value is supported by all browsers targeted as the minimum by TailwindCSS v4. Since DaisyUI also relies on TailwindCSS v4, I believe using `1lh` is justified. Moreover, if it's not supported, the fallback value of `3.75rem` still works well.

### DIFF
--- a/packages/daisyui/src/components/collapse.css
+++ b/packages/daisyui/src/components/collapse.css
@@ -120,12 +120,8 @@
     width: 100%;
     padding: 1rem;
     padding-inline-end: 3rem;
-    min-height: 3.75rem;
+    min-height: 1lh;
     transition: background-color 0.2s ease-out;
-
-    @supports (min-height: 1lh) {
-      min-height: 1lh;
-    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks!

PR #3899, opened 2025-05-21
Add a note regarding theme-controller & prefersdark

ItsJustMeChris:
I don't think it was immediately clear to some people (me included) why setting data-theme might not have been used by daisy while using `prefersdark`, this adds a note to address / raise awareness to how to get data-theme and system defaults to work together nicely.

### DIFF
--- a/packages/docs/src/routes/(routes)/docs/themes/+page.md
+++ b/packages/docs/src/routes/(routes)/docs/themes/+page.md
@@ -25,6 +25,8 @@ You can manage themes by adding brackets in front of `@plugin "daisyui"` in your
 - You can set `--default` flag for a theme to make it the default theme.
 - You can also set `--prefersdark` flag for a theme to make it the default theme for dark mode (prefers-color-scheme: dark).
 
+> When utilizing `--prefersdark`, be sure to include a [theme‑controller](/components/theme-controller/). Without it, `<html data-theme>` will be ignored, and the preferred system theme will always be used instead.
+
 ## Enable a built-in theme
 
 By default, `light` and `dark` themes are enabled. Let's enable `cupcake` theme:  


saadeghi:
Hey, thanks for the PR.  
The note you added is not true. Using `theme-controller` is not required.  

Here's an example: https://play.tailwindcss.com/GGLeiZHkQ8  

https://github.com/user-attachments/assets/375c23c7-3dd9-43cf-a723-9da9cb19e343

If it's not working correctly in your project, please open a [discussion](https://github.com/saadeghi/daisyui/discussions) and provide a reproduction link so I can help finding the issue.

ItsJustMeChris:
that's for a single element, if the data-theme is applied to the <html> root specifically it will not adhere to the theme unless a theme controller is present. 
MTOqq8IkkmWM3HSh@2x
Tr5LBdMamwwL48pC@2x

Adding theme controller to App.vue causes the theme to be respected

saadeghi:
Can you share your CSS file? (only the daisyUI parts)

ItsJustMeChris:
Doing that causes an FOUC to the system theme even if seeded with a cookie in SSR

```code
          innerHTML: `(function(){
              function getCookie(name) {
                const match = document.cookie.match(new RegExp('(^| )' + name + '=([^;]+)'));
                if (match) return match[2];
              }
              const theme = getCookie('theme');
              console.log('Theme cookie:', theme);
              if(theme === 'light-alpha'){
                document.documentElement.setAttribute('data-theme','light-alpha');
              } else if(theme === 'dark-alpha'){
                document.documentElement.setAttribute('data-theme','dark-alpha');
              } else if(theme === 'system'){
                document.documentElement.removeAttribute('data-theme');
              }
              const supportedLangs = ['en','es','fr'];
              let lang = getCookie('language');
              if(!lang){
                const nav = navigator.languages?.[0] || navigator.language;
                const short = nav && nav.split('-')[0];
                if(short && supportedLangs.includes(short)){
                  lang = short;
                }
              }
              if(lang){
                document.documentElement.setAttribute('lang', lang);
              }
            })();`,
          type: "text/javascript",
```

the code that was shared in that sample does not cause an flash of inconsistent styles when using the theme controller in place - though, outside of this, `moving dark above light does _remove_ the need for the theme controller` but `it causes inconsistent styling flash`.

saadeghi:
If the server is rendering the page with correct `data-theme`, there should be no flash of wrong theme.  
Because CSS is instant and there's no execution time (unless you're leading the CSS file using JS)  
If the `data-theme` is being set on the browser using JS, you may want to do that before page loads, otherwise page will load with the default theme then JS sets the correct `data-theme` which will cause a flash.

ItsJustMeChris:
Yep you're right that code that was supposed to seed ssr was wrong. Sorry 😓 thanks for the back and forth anyways. Buying a hat now

saadeghi:
No problem 💚🙏

PR #3902, opened 2025-05-23
Remove theme-change from +page.md

OskarD:
It's a bad recommendation, as there is no documentation how to use it with daisyUI


saadeghi:
It's not a integration.  
- daisyUI themes work with `data-theme` attribute and is agnostic about how you setup JS.   
- `theme-change` package reads and sets data-theme value to local storage.
So they work together without any setup.

If it's not working for you, please open a discussion and provide a minimal reproduction link so I can help.

OskarD:
It's not just me: https://github.com/saadeghi/theme-change/issues/30

OskarD:
Yes

saadeghi:
I will make a next.js example repo this Monday

PR #3932, opened 2025-06-01
fix: increase drawer z-index to prevent overlap issues. closes: #3929

### LINKED ISSUE #3929: bug: Floating Label overlays the drawer component

https://github.com/kensodemann/show-floating-labels

5.0.43

All browsers

### Describe your issue

I have an application that has some forms using floating labels. I also have a drawer menu to the left. On smaller form factors, this drawer is opened via a hamburger menu. If this takes place while there is data in the inputs, and thus the labels are floating, they float above the drawer.

I have reproduced this in a simple app. To replicate, run the (see instructions in README.md file), fill in some of the inputs so at least one of the labels floats, then open the drawer using the hamburger menu. You should see the floating labels float above the menu.

Inputs with the labels floating:
Image

Labels floating over the drawer:
Image

I noticed the floating labels use a `z-index` of `2`, and removing that solves _this_ issue, but I have no idea if it causes other issues.

```css
.floating-label {
    &:focus-within, &:not(:has(input:placeholder-shown, textarea:placeholder-shown)) {
        > span {
            opacity: 100%;
            top: 0;
            translate: -12.5% calc(-50% - 0.125em);
            scale: 0.75;
            pointer-events: auto;
            z-index: 2;
        }
    }
}
```

sulimanbenhalim:
## Problem
Fixes #3929

Drawer component appears behind floating-label and possibly other component with z-index more than 1 this is causing overlay and visibility issues in certain layouts

## Fix
Increased drawer z-index from `z-1` to `z-10` for proper layering and prevent the drawer from being hidden behind other page elements like what happened with floating-label.

## Testing
Verified that drawer now properly appears above floating-label

### DIFF
--- a/packages/daisyui/src/components/drawer.css
+++ b/packages/daisyui/src/components/drawer.css
@@ -9,7 +9,7 @@
 }
 
 .drawer-side {
-  @apply pointer-events-none invisible fixed start-0 top-0 z-1 col-start-1 row-start-1 grid w-full grid-cols-1 grid-rows-1 items-start justify-items-start overflow-x-hidden overflow-y-hidden overscroll-contain opacity-0;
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you!

PR #3939, opened 2025-06-04
Color SVG fills the base content color by default

sabslikesobs:
DaisyUI sets "color" to --color-base-content on all elements. In addition to all text, this colors SVGs, but only the stroke part, not the fill. The SVG fill in icons, like an icon from rails_icons (https://github.com/Rails-Designer/rails_icons) isn't colored.

This can be surprising when injecting icons anywhere in buttons or input forms, for example. I was working on a password form, and a path-only icon was colored correctly, but a fill-only icon wasn't, and I needed to set the color myself.

It might actually be a good idea to set "fill" whenever "color" gets set, too... but I figured I'd bring up the idea here.

### DIFF
--- a/packages/daisyui/src/base/rootcolor.css
+++ b/packages/daisyui/src/base/rootcolor.css
@@ -2,4 +2,5 @@
 [data-theme] {
   background-color: var(--root-bg, var(--color-base-100));
   color: var(--color-base-content);
+  fill: var(--color-base-content);
 }


sabslikesobs:
> It might actually be a good idea to set "fill" whenever "color" gets set, too... but I figured I'd bring up the idea here.

Here's an example of what I mean:

Screenshot From 2025-06-05 10-17-17

The top one is an in-HTML SVG inside an `alert alert-soft alert-info`, without any `fill-` styling. The stroke is the correct color, but the circle is not. For that, I have to add `fill-info` somewhere in the div. It would be pretty easy to smooth this UX by setting `fill` wherever `color` is also set.

saadeghi:
Hey, Thanks for the PR.  
I don't think it's a good idea to set fill color by default.  
SVGs may or may not need a fill color OR stroke color.  

For example you can compare "Outline" icons and "Solid" icons at https://heroicons.com/  
If we set a default fill color, all "Outline" icons will inherit a fill color but they do not need it at all. They only need stroke colors.  
Meanwhile "Solid" icons do not need stroke color and they need fill color instead.  

If you want your "Outline" icons to inherit their stroke color from the default text color, you should add `stroke="currentColor"` to the SVG,  
and if you want your "Solid" icons to inherit their fill color from the default text color, you should add `fill="currentColor"` to the SVG

sabslikesobs:
Yes, you are right. I hadn't thought of that, and had assumed SVGs used "color" and not "stroke"... I guess it's up to the icon library on how they do it.

For my case, I fixed the general behavior of icons with

```
:root {
  fill: currentColor;
}
```

PR #3948, opened 2025-06-08
Fix .css file name

gurbaxani:


saadeghi:
Why?

gurbaxani:
That's the file name that the SvelteKit CLI generates by default. So no point creating a style.css when app.css exists, hence the suggestion.

saadeghi:
I see.  
Initially I only tested `SvelteKit minimal` template which doesn't have a CSS file.  
I think you used `SvelteKit demo` template which has a `app.css` already.

Can you also update the file name on line `47` so I can merge?

gurbaxani:
Good catch, fixed both imports. Thank you for this useful Tailwind plugin!

PR #3971, opened 2025-06-27
feat: add custom color picking and hex code input to the theme generator

SamuelLHuber:
addresses https://github.com/saadeghi/daisyui/discussions/3970

### DIFF
--- a/packages/docs/src/components/ColorPalette.svelte
+++ b/packages/docs/src/components/ColorPalette.svelte
@@ -1,6 +1,7 @@
 <script>
   import ContrastMeter from "$components/themegenerator/ContrastMeter.svelte"
   import { validateColor } from "$lib/themeGeneratorValidation"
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks for the PR.  
There are some parts that I need to change.  
1. The default browser color picker is RGB and doesn't allow P3 colors so I don't think it's a good idea to use that. Also dragging colors in the color picker is laggy because the URL of the page changes live based on the chosen color.
2. the layout doesn't look good
3. the conversion is doesn't always work and is values are not validated. for example if you write `red` in custom color field, it shows `NaN` which is invalid.

saadeghi:
Added OKLCH, HSL, RGB and Hex color pickers ✅  
Thanks for the idea.  
Screenshot 2025-09-26 at 2 23 06 PM

PR #3987, opened 2025-07-09
Refactor: Use transparent keyword for background-color in glass.css

Mirza-Samad-Ahmed-Baig:
This PR refactors the background-color property in glass.css to use the transparent keyword instead of #0000. While #0000 is functionally equivalent, transparent is more explicit, readable, and idiomatic in CSS. This change improves code clarity and maintainability without altering the visual output.


saadeghi:
Hey, thanks for the PR but this doesn't fix any issue.  
This property is not something a daisyUI user needs to interact with.  
The reason `#0000` is used everywhere instead of `transparent` is to maximize the compressed size of CSS file.

PR #3992, opened 2025-07-12
fix: apply theme-aware hover styling to Cally day buttons. closes: #3991

### LINKED ISSUE #3991: bug: Hovering over days of Cally calendar does not use correct theme color

https://daisyui.com/components/calendar/#cally-calendar-example

5.0.46

All browsers

### Describe your issue

When I hover over the day buttons on the Cally component, the background color doesn't match the theme color.  Instead, it is using the Cally default theming, which sets the background to `rgba(0, 0, 0, 0.05)`.  In contrast, hovering over the prev/next buttons sets the background to `var(--color-base-200)`.  In many themes these are similar, but here is a screenshot from the docs with the `valentine` theme where I have used devtools to simultaneously set the "next" button and "6" to hover.  Notice that the colors are different:

Image

The current selector in calendar.css is incorrectly only applying to the heading button.

I have a fix for this and can submit a PR if you want.

sulimanbenhalim:
## Problem
Fixes #3991

Cally calendar day buttons were using the component's default hover background (`rgba(0, 0, 0, 0.05)`) instead of the theme-aware daisyUI color. This created visual inconsistency where navigation buttons (prev/next) correctly used `var(--color-base-200)` on hover, but day buttons did not match the theme colors.

## Fix
Added missing `::part(day):hover` CSS selector with `background: var(--color-base-200)` to ensure day buttons use the same theme-aware hover styling as navigation buttons.

## Testing
Verified that day buttons now properly match the theme hover color across different daisyUI themes in the playground.

### DIFF
--- a/packages/daisyui/src/components/calendar.css
+++ b/packages/daisyui/src/components/calendar.css
@@ -29,6 +29,9 @@
     border-radius: var(--radius-field);
     font-size: 0.7rem;
   }
+  ::part(day):hover {
+  background: var(--color-base-200);
+  }
   ::part(button day today) {
     background: var(--color-primary);
     color: var(--color-primary-content);


saadeghi:
Thanks

PR #3993, opened 2025-07-12
fix: react-day-picker disableNavigation not styling properly. closes: #3990

### LINKED ISSUE #3990: bug: react-day-picker fails to set proper color or opacity for `disableNavigation`

https://stackblitz.com/edit/daisyui5-6thuepq1?file=src%2FApp.jsx

v5.0.22

All browsers

### Describe your issue

In `react-day-picker` with DaisyUI styling 

    <DayPicker mode="single" className="react-day-picker"></DayPicker>

if you add the attribute `disableNavigation`, the controls *aren't* greyed out but should be. (see stackblitz link. only runs on chrome, not FF, idk why)

For reference, this is what the default styles provided by rdp do (you're supposed to *not* include them as per daisy docs):

```css
.rdp-button_next:disabled, .rdp-button_next[aria-disabled="true"], .rdp-button_previous:disabled, .rdp-button_previous[aria-disabled="true"] {
  cursor: revert;
  opacity: var(--rdp-nav_button-disabled-opacity);
}
```

sulimanbenhalim:
## Problem
Fixes #3990

When using `react-day-picker` with the `disableNavigation` prop, the navigation buttons (previous/next) are not properly styled as disabled. They remain clickable-looking instead of being greyed out.

This happens because `react-day-picker` uses `aria-disabled="true"` instead of the actual `disabled` attribute, but daisyUI's CSS only targets `:disabled`.

## Fix
Added CSS selectors to handle both `:disabled` and `[aria-disabled="true"]` states for navigation buttons:

- Added `&[aria-disabled="true"]` selector alongside existing `&:disabled`
- Disabled hover effects for disabled navigation buttons

## Testing
Created test page in playground comparing normal vs disabled navigation

### DIFF
--- a/packages/daisyui/src/components/calendar.css
+++ b/packages/daisyui/src/components/calendar.css
@@ -139,9 +139,16 @@
     padding: 0;
     display: inline-flex;
     position: relative;
-    &:disabled {
+
+    &:disabled,
+    &[aria-disabled="true"] {
       cursor: revert;
       opacity: 0.5;
+    }
+
+    &:disabled:hover,
+    &[aria-disabled="true"]:hover {
+      background-color: transparent;
     }
   }
   .rdp-chevron {


saadeghi:
Thank you!

PR #3994, opened 2025-07-12
fix: use semantic fallbacks for checkbox states, improve indeterminate visibility. closes: #3979

### LINKED ISSUE #3979: bug: Missing background color for indeterminate checkboxes

https://stackblitz.com/edit/daisyui5-lhngeyg5?file=src%2Fmain.js%3AL19

v5.0.43

Edge, Firefox

### Describe your issue

Indeterminate checkboxes with a checkbox-xxx class for color (i.e. checkbox-primary) have the dash almost invisible for some colors. It happens to different colors for light and dark theme.
It seems to happen when the --color-xxx-content color is close to the screen background (base-content?).
Checked checkboxes are good because there is a rule for '.checkbox:checked, .checkbox[aria-checked="true"]' that gives 'background-color: var(--input-color);'

I believe this can be fixed in src/components/checkbox.css by copying line 40 "background-color: var(--input-color, #0000);" and inserting it after line 64 "&:indeterminate {". This is quick, so if you just do it without assigning me to the task and making me create a pull request, I'm fine with that.

I don't have a stackblitz account, so in case the link isn't valid, here is the code: [stackblitz.txt](https://github.com/user-attachments/files/21024279/stackblitz.txt)

Thanks for daisyUI! It does a much better job than I could have done myself.

sulimanbenhalim:
## Problem
Fixes #3979

Indeterminate checkboxes with color variants (checkbox-primary, checkbox-secondary, etc.) have nearly invisible dash marks due to poor contrast. The issue occurs when color content is similar to the background, making the dash blend invisibly.

Additionally, the codebase has an inconsistency where borders use semantic fallbacks (`color-mix(in oklab, var(--color-base-content) 20%, #0000)`) but the checked state uses a hardcoded `#0000` fallback.

## Fix
Applied a solution using semantic fallbacks consistent with daisyUI's established patterns:

**1. Updated checked state background-color:**
```css
// Before
background-color: var(--input-color, #0000);

// After  
background-color: var(--input-color, color-mix(in oklab, var(--color-base-content) 20%, #0000));
```

**2. Added indeterminate state background-color:**
```css
&:indeterminate {
  background-color: var(--input-color, color-mix(in oklab, var(--color-base-content) 20%, #0000));
  &:before { ...
```

## Testing
Verified all color variants (primary, secondary, accent, neutral, info, success, warning, error) in playground

### DIFF
--- a/packages/daisyui/src/components/checkbox.css
+++ b/packages/daisyui/src/components/checkbox.css
@@ -37,7 +37,7 @@
 
   &:checked,
   &[aria-checked="true"] {
-    background-color: var(--input-color, #0000);
+    background-color: var(--input-color, color-mix(in oklab, var(--color-base-content) 20%, #0000));
     box-shadow:
       0 0 #0000 inset,
       0 8px 0 -4px oklch(100% 0 0 / calc(var(--depth) * 0.1)) inset,
@@ -62,6 +62,7 @@
   }
 
   &:indeterminate {
+    background-color: var(--input-color, color-mix(in oklab, var(--color-base-content) 20%, #0000));
     &:before {
       @apply rotate-0 opacity-100;
       translate: 0 -35%;


saadeghi:
Thank you!

PR #3998, opened 2025-07-12
fix: markdown rendering issue with backticks in component headings. closes: #3996

### LINKED ISSUE #3996: docs: Markdown rendering issue on collapse component documentation page

https://daisyui.com/components/collapse/?lang=en#custom-colors-for-collapse-that-works-with-focus

### Describe the issue

Hi, I noticed a rendering issue on the [collapse component documentation page](https://daisyui.com/components/collapse/?lang=en#custom-colors-for-collapse-that-works-with-focus). The page only displays the text:

> "Use Tailwind CSS"

However, when I inspected the source, the actual markdown is:

```md
#### Use Tailwind CSS `group` and `group-focus` utilities to apply style when parent div is focused
```

It seems that the use of backticks might be causing the content to be cut off or improperly rendered. Could this be a markdown parsing issue on the site? Would appreciate if this could be fixed to make the documentation clearer.

Thanks for your great work on DaisyUI!

Image

Image

Chrome

sulimanbenhalim:
## Problem
Fixes #3996

Headings in component docs were losing their markdown formatting. For example, the collapse page heading "Use Tailwind CSS `group` and `group-focus` utilities..." was showing up as plain text without the part where backticks starts around `group` and `group-focus`.

## Fix
The original `getHeadingText()` function was only grabbing the first text node and ignoring everything else, rewrote it to handle all the different node types properly.

Added a `nodesToHtml()` function that converts the markdown AST nodes into actual HTML - so `inlineCode` becomes `<code>`, `emphasis` becomes `<em>`, etc. Also had to fix some quote escaping issues where href attributes were breaking because of nested quotes.

Changed the component structure to store both the plain text (for IDs) and the original nodes (for rendering), then updated Component.svelte to detect when there's HTML content and render it instead of just treating everything as plain text.

## Testing
Checked the collapse docs page where the issue was reported and the heading now shows up correctly with proper code styling, see below:

image

Added a test case with different markdown combinations to make sure bold, italic, code, and links all work together. Built and tested in the docs to verify everything renders properly:

image

### DIFF
--- a/packages/docs/src/components/Component.svelte
+++ b/packages/docs/src/components/Component.svelte
@@ -85,7 +85,11 @@
   {/if}
   {#if desc}
     <div class="pb-4 text-xs opacity-70">
-      {$t(desc)}
+      {#if desc.includes('<')}
+        {@html desc}
+      {:else}
+        {$t(desc)}
+      {/if}
     </div>
   {/if}
   <div class="tabs tabs-lift">
--- a/packages/docs/src/lib/mdsvex/remark-render-component.js
+++ b/packages/docs/src/lib/mdsvex/remark-render-component.js
@@ -1,10 +1,44 @@
 import { visit } from "unist-util-visit"
 
+// Helper to escape quotes for HTML attributes
+function escapeQuotes(text) {
+  return text.replace(/"/g, "&quot;").replace(/'/g, "&#39;")
+}
+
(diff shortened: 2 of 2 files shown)


saadeghi:
Thank you @sulimanbenhalim 🙏

PR #4008, opened 2025-07-18
chore(docs-pt): adjust navbar translation

horberlan:
adjusts the translation of the name of the “navbar” component, which was in a free translation, creating ambiguity in its description.


saadeghi:
Thanks!

Issue #4021, opened 2025-07-25
bug: disabling a fieldset does not apply correct styles for nested labels with inputs


chenzhekl:
### Reproduction URL (Required)

https://play.tailwindcss.com/CVhpVvyYAx

### What version of daisyUI are you using?

v5.0.46

### Which browsers are you seeing the problem on?

Chrome

### Describe your issue

As the title said, disabling a fieldset did not apply correct styles for nested labels with inputs. They look exactly the same as being enabled but are not editable.

jhobz:
I'm also experiencing this issue with select elements.

<img width="427" height="202" alt="Image" src="https://github.com/user-attachments/assets/257d083a-ec4a-4e85-8eed-d3f0d86ae84f" />

Note the range, number input, and button are all properly styled, but the select elements are not.

PR #4038, opened 2025-08-03
input: update styles for disabled state in forms and fieldsets

yoshi6jp:
Fix input disabled state when parent fieldset or form is disabled

This PR fixes issue #4021 where the `.input` component doesn't show the disabled state when its parent `<fieldset>` or `<form>` element has the `disabled` attribute.

**Changes:**
- Added `fieldset:disabled &` and `form:disabled &` selectors to the disabled state rules
- Now the input will properly display as disabled when:
  - The input element itself has the `disabled` attribute
  - The `.input` element has the `disabled` attribute
  - The parent `<fieldset>` has the `disabled` attribute
  - The parent `<form>` has the `disabled` attribute

This follows HTML semantics where form elements inside a disabled fieldset or form are automatically disabled.

### DIFF
--- a/packages/daisyui/src/components/select.css
+++ b/packages/daisyui/src/components/select.css
@@ -58,7 +58,8 @@
   }
 
   &:has(> select[disabled]),
-  &:is(:disabled, [disabled]) {
+  &:is(:disabled, [disabled]),
+  fieldset:disabled & {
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you!

PR #4053, opened 2025-08-14
fix: steps component text overlap on mobile. closes: #4043

### LINKED ISSUE #4043: bug: Steps with long texts, crashing between them

https://play.tailwindcss.com/L2yalFmJcY

v5.0.50

All browsers

### Describe your issue

Using the following example code:

```
<ul class="steps my-4 w-full">
  <li class="step step-primary">Creación de cuenta</li>
  <li class="step step-primary">Configuración de proyecto</li>
  <li class="step">Prueba gratuita</li>
  <li class="step">Método de pago</li>
  <li class="step">Checkout</li>
  <li class="step">Activation</li>
</ul>
```

when looking at mobile view the text inside the steps is crashing between them causing a weird UI experience. 

Any help to solve it is appreciate it. 

**Web View:**
Image

**Mobile View**
Image

sulimanbenhalim:
## Problem
Fixes #4043

The steps component text overlaps on mobile devices when using long text content. This occurs because `grid-auto-columns: 1fr` forces equal width distribution, causing text to overflow horizontally into other steps.

## Fix

- added `overflow-wrap: break-word`, `word-break: break-word`, and `hyphens: auto`
- changed font size to `0.75rem` and min-width to `3rem` on screens ≤768px  
- changed padding to `0 0.25rem` for text fit

## Testing
Tested in playground with original Spanish text example and extended English text:
image
image

However it's better to use vertical steps on mobile screen anyways

### DIFF
--- a/packages/daisyui/src/components/steps.css
+++ b/packages/daisyui/src/components/steps.css
@@ -10,6 +10,16 @@
     min-width: 4rem;
     --step-bg: var(--color-base-300);
     --step-fg: var(--color-base-content);
+
+    @media (max-width: 768px) {
+      min-width: 3rem;
+      font-size: 0.75rem;
+      overflow-wrap: break-word;
+      word-break: break-word;
+      hyphens: auto;
+      padding: 0 0.25rem;
+    }
+
     &:before {
       @apply top-0 col-start-1 row-start-1 h-2 w-full;
       border: 1px solid;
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks for the effort @sulimanbenhalim but I can't merge this unfortunately. This is expected behavior. when steps is horizontal, it will take space as much as the content.  
As you mentioned, for mobile, they can use a vertical step. Or they can make it scrollable or they can decrease the font size based on the need. Examples: https://github.com/saadeghi/daisyui/issues/4043#issuecomment-3189373356

But it's not a good idea to change the font size, width or paddings just to force it fit there.  
This is a breaking change for apps that are relying on current behavior.  
For example, if I already customized my step to have specific font size, or width or padding in mobile, this PR will reset my customization. We can't do such things in a patch version like `5.0.x` it's a breaking change that we can do in `6.0.0` (if we have to)

PR #4056, opened 2025-08-14
fix: improve text contrast for selected cally dates on hover closes: #4042

### LINKED ISSUE #4042: bug: cally hover on 'selected' dates is hard to read

https://daisyui.com/components/calendar/#cally-calendar-example

5.0.50

All browsers

### Describe your issue

Hovering your mouse over `cally` dates that have the `.selected` class are hard to read. This is viewable straight in the docs

Image

Image

Image

Image

Is there any way to apply the `text-base-content` class or something similar when a cally `.selected` is hovered so that the text is still readable? I double checked and it's the same way for `<date-range>`. All of the items in the range have `.selected` and on hover it's hard to read the text.

sulimanbenhalim:
## Problem
Fixes #4042

Hovering over selected dates in cally calendars makes text hard to read due to poor contrast
Selected dates have light text (var(--color-base-100)) that becomes unreadable when the hover state applies a light background (var(--color-base-200)).

## Fix
Added specific CSS rule for selected dates on hover that changes text color to var(--color-base-content) while keeping the var(--color-base-200) background

## Testing
Tested in playground with both single date and date range calendars

### DIFF
--- a/packages/daisyui/src/components/calendar.css
+++ b/packages/daisyui/src/components/calendar.css
@@ -30,7 +30,7 @@
     font-size: 0.7rem;
   }
   ::part(day):hover {
-  background: var(--color-base-200);
+    background: var(--color-base-200);
   }
   ::part(button day today) {
     background: var(--color-primary);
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you!

PR #4067, opened 2025-08-20
feat(a11y): add prefers-reduced-motion support

abh:
Disable animations when users prefer reduced motion:

- skeleton: remove shimmer animation
- progress: use solid fill instead of moving stripes
- collapse: disable grid transitions and rotations
- toast, dropdown, rating, radio: disable entry animations
- swap, tooltip: disable transform transitions
- carousel: disable smooth scroll (already implemented)

Improves accessibility for users with vestibular disorders

NOTE: I wasn't able to test this because the current animations
makes me feel sick, so I was making these changes just by
code inspection.

### DIFF
--- a/packages/daisyui/src/components/carousel.css
+++ b/packages/daisyui/src/components/carousel.css
@@ -4,6 +4,10 @@
   scroll-behavior: smooth;
   scrollbar-width: none;
 
+  @media (prefers-reduced-motion: reduce) {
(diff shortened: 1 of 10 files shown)


saadeghi:
Thanks for the PR.  
I think it would be cleaner to put the original animations and transition in media query rather than adding an override.  
I will add a commit to this PR soon 👍

PR #4097, opened 2025-09-06
fix: accordian examples to use semantic html

81reap:
update accordian examples to use semantic html

### DIFF
--- a/packages/docs/src/routes/(routes)/components/accordion/+page.md
+++ b/packages/docs/src/routes/(routes)/components/accordion/+page.md
@@ -39,110 +39,110 @@ classnames:
 
 
 ### ~Accordion using radio inputs
-<div class="collapse bg-base-100 border border-base-300">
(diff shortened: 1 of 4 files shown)


saadeghi:
Hey, Thanks for the PR!  
I saw the conversation on Twitter but there's a misunderstanding here.  
[Collapse](https://daisyui.com/components/collapse/#-collapse-using-details-and-summary-tag) uses `<details>` and `<summary>` because each item can and should be open independently.  
[Accordion](https://daisyui.com/components/accordion/) is different. Only one item must be open and opening an item must close others. This requires radio inputs to store the state.

Elements can be `<section>` or `<h*>` tags instead of `<div>` but it can't be `<details>` and `<summary>` as these tags can only open/close independently in HTML.

See this example:  
https://play.tailwindcss.com/a6ZlwPjI5y

81reap:
Ah, that's what I get for whipping out a PR at 02:00. Thanks for the explanation! :)

PR #4111, opened 2025-09-10
fix: Firefox Noise Effect

### LINKED ISSUE #3583: bug: Noise Option Broken on Firefox

v5.0.0

Firefox

### Reproduction URL

https://daisyui.com/theme-generator/

### Describe your issue

The noise effect does not work properly on Firefox when compared to Chrome and Safari, only displaying in the corner rather than over the entire element.

Image

bert-mccutchen:
## Description

Fixes #3583

> The noise effect does not work properly on Firefox when compared to Chrome and Safari, only displaying in the corner rather than over the entire element.

## Changes

1. Added a `viewBox` to fix the `--fx-noise` SVG rendering on Firefox.
2. Changed width and height on the `rect` to be static. This fixes rendering on Chromium after adding the `viewBox`.

### Firefox

| Before | After |
|---|---|
| firefox-before | firefox-after |

### Chromium

| Before | After |
|---|---|
| chromium-before | chromium-after |


saadeghi:
Thank you!

ItsJustMeChris:
Thank you 🫡

PR #4121, opened 2025-09-14
fix: sync margin inline start of select in label after change of padding inline start (ref #4114)

### LINKED ISSUE #4114: bug: Grid + Label + Select border left bug  and appearance-none not working with label+select

https://play.tailwindcss.com/xGLwNMovro

v5.1.10

Chrome

### Describe your issue

1. grid + label + select  = select left border bug (chrome & firefox)
2. appearance-none + label + select = not working as expected (chrome only)

pdanpdan:
After the change of padding-inline start of .select  from ps-4 to ps-3 in 449378e93adba1fb71aba2c26515f4b9643244f9 the margin-inline-start was forgotten.

### DIFF
--- a/packages/daisyui/src/components/select.css
+++ b/packages/daisyui/src/components/select.css
@@ -32,7 +32,7 @@
       calc(0% + 16px) calc(1px + 50%);
   }
   select {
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you 🙏

PR #4125, opened 2025-09-18
Update CDN page

steve02081504:


saadeghi:
Hey, what are the things that doesn't work with `@tailwindcss/browser` dynamically?

steve02081504:
> Hey, what are the things that doesn't work with `@tailwindcss/browser` dynamically?

I can't give a minimum repro exactly, but I did have a related issue in fount development, history:
https://discord.com/channels/1204154732978118716/1205266588371259492/1394891051319300229
image

saadeghi:
Thanks.  
People can use Uno CSS or any other library for utility classes but I'm not sure if suggesting it as a replacement for Tailwind CSS would be a good idea if the issue is a bug.  
I think the bug must be reported to Tailwind CSS repo and they will fix it.  

About the opacity values, I added the note in the related section in docs:  
https://daisyui.com/docs/colors/#opacity-values

PR #4134, opened 2025-09-24
docs: Removed out of scope text from codeblock

TheWilley:
When copying the JSX code from the "[Helper Dropdown](https://daisyui.com/components/dropdown/#helper-dropdown)" section, the text "A normal text and a helper dropdown" is incorrectly added at the top, causing a syntax error when pasted. This pull request simply removes that text.


saadeghi:
Thank you 🙏

PR #4142, opened 2025-10-01
fix(select): prevent line breaks on long text

### LINKED ISSUE #4105: bug: Select component wraps long option labels

https://play.tailwindcss.com/rPGJ7bjVLG

v5.1.9

Chrome

### Describe your issue

Long select options now wrap in v5.1.9. If this was an intentional design change, it should probably be documented on the Select component in the docs. Otherwise, it is a regression.

ermesonqueiroz:
## Description

Fixes #4105 

> Long select options now wrap in v5.1.9. If this was an intentional design change, it should probably be documented on the Select component in the docs. Otherwise, it is a regression.

## Changes

- Truncate text in `<select>`
- Changed the options container width to same width of `<select>`
- Truncate text in `<option>`

## Screenshots (Chrome)
| Before | After|
| --- | --- |
| Screenshot 2025-09-30 at 22 54 48 | Screenshot 2025-09-30 at 22 55 32 |

## Considerations

I initially attempted to implement `text-overflow: ellipsis` to add a `...` indicator for truncated text. However, to make `text-overflow: ellipsis` work, I needed to use `-webkit-appearance: none;`.

This approach caused a significant side-effect: it stripped the custom styling from the `<option>` elements, making them revert to the default OS appearance. To avoid this styling regression and preserve the component's design, I opted for the current solution of hiding the overflow without an ellipsis.

### DIFF
--- a/packages/daisyui/src/components/select.css
+++ b/packages/daisyui/src/components/select.css
@@ -19,6 +19,8 @@
     4px 4px,
     4px 4px;
   background-repeat: no-repeat;
+  white-space: nowrap;
+  overflow: hidden;
   text-overflow: ellipsis;
   box-shadow:
     0 1px color-mix(in oklab, var(--input-color) calc(var(--depth) * 10%), #0000) inset,
@@ -79,6 +81,7 @@
     &::picker(select) {
       color: inherit;
       max-height: min(24rem, 70dvh);
+      max-width: anchor-size(width);
       border: var(--border) solid var(--color-base-200);
       @apply rounded-box my-2 p-2;
       background-color: inherit;
@@ -111,6 +114,8 @@
       transition-property: color, background-color;
       transition-duration: 0.2s;
       transition-timing-function: cubic-bezier(0, 0, 0.2, 1);
+      white-space: nowrap;
+      overflow: hidden;
       &:not(:disabled) {
         &:hover,
         &:focus-visible {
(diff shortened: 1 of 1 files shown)


SanmayJoshi:
If the (long) text could marquee on hover (for select as well as option), it would be amazing. Or an option via an additional class may also be ok.

This should probably not stop the PR though, especially if implementing this is significant additional work.

saadeghi:
> If the (long) text could marquee on hover (for select as well as option), it would be amazing.

That's a solution for a problem that shouldn't exist in the first place. If the text is long, use a shorter text instead of torturing the user 😅

pdanpdan:
I think this solution breaks a lot of things:
- the width of the select
- it should not force the options to be single line

I would go with styling default `selectedcontent` and adding in the documentation examples with `<button><selectedcontent></selectedcontent></button>`

https://play.tailwindcss.com/f1azii5wSq?file=css

saadeghi:
@pdanpdan's concerns are valid. if there's no way to read the text, it's a bad UX.  
I added some changes.  
Before this PR:  
Text was overflowing the page
Screenshot 2025-10-03 at 11 01 51 PM

after @ermesonqueiroz changes:  
Screenshot 2025-10-03 at 10 45 32 PM

after my additional changes:  
Screenshot 2025-10-03 at 11 00 45 PM

Now: 
- text is one-line when select is not open
- text is multi line (if needed) and readable when select is open and it wont overflow the page
- select dropdown doesn't have a strict width. it grows only if needed.

PR #4152, opened 2025-10-05
docs: fix code example differs from the actual HTML code

Flo0807:
I noticed that the code for "Table with pinned-rows and pinned-cols" in the table documentation differs from the HTML code displayed on the page. The fixed width and height are important for the `position: sticky;` (used by `table-pin-rows` and `table-pin-cols`) to work properly.


saadeghi:
Thank you!

PR #4153, opened 2025-10-05
Add docs for ember

NullVoxPopuli:
wip

### DIFF
--- a/packages/docs/src/routes/(routes)/docs/install/ember/+page.md
+++ b/packages/docs/src/routes/(routes)/docs/install/ember/+page.md
@@ -1,0 +1,61 @@
+---
+title: Install daisyUI for Ember
+desc: How to install Tailwind CSS and daisyUI in a Ember project
+---
+
+<script>
+  import Translate from "$components/Translate.svelte"
+</script>
+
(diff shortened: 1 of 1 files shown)


NullVoxPopuli:
I thiiink all I have left is to add a link to this page: https://daisyui.com/docs/install/ and add the icon?

NullVoxPopuli:
actually, where is that?
I see there is some frameworks.yaml, but I can't find it?

```
    const response = await fetch(`${PUBLIC_DAISYUI_API_PATH}/data/frameworks.yaml`)
```

saadeghi:
Thank you 🙏

> I thiiink all I have left is to add a link to this page: https://daisyui.com/docs/install/ and add the icon?

> I see there is some frameworks.yaml, but I can't find it?

I will do that 👍  
File is here https://github.com/daisyui/api/blob/master/docs/data/frameworks.yaml I will update.

NullVoxPopuli:
> I will do that 

Thank you!!

This is the icon to use: https://emberjs.com/logos/
image


It's also in some iconify sets

saadeghi:
Thank you @NullVoxPopuli 💚

PR #4160, opened 2025-10-10
fix(collapse): fix tailwind default .collapse with revert-layer instead of visible

pdanpdan:
Setting `.collapse` to `visibility: visible` creates lots of problems (one such example is in modal)

Another alternative is to set `visibility: unset`


saadeghi:
Thank you!

Issue #3835, opened 2025-05-01
bug: Collapse inside Bottom Modal glitches the open animation


brunocalou:
### Reproduction URL (Required)

https://codepen.io/brunocalou/pen/vEEpyWw

### What version of daisyUI are you using?

5.0.32

### Which browsers are you seeing the problem on?

All browsers

### Describe your issue

When collapse component is rendered inside a bottom modal, the open animation doesn't work properly.

https://github.com/user-attachments/assets/8d247007-3e1b-43ab-9b45-b49cdb8d4768

https://github.com/user-attachments/assets/9bb5a267-579f-4d81-96e1-b8139f24fddf

I've investigated a little bit and the `visibility` property on `collapse` and `collapse-content` is causing this issue. I don't know why exactly, but for now, overriding it to `unset` fixes the issue.

```html
<div class="collapse [visibility:unset]!">
  <div class="collapse-content [visibility:unset]!">
    ....
  </div>
</div>
```

https://github.com/user-attachments/assets/e7ee444d-0a5e-4ad8-a947-9cd10f798378

SebastienFouchard:
@brunocalou example clearly reproduces the issue, but the provided
screenshots don’t fully capture the animation glitch.  
Here is a new screenshot that better illustrates the problem.

https://github.com/user-attachments/assets/9141a64c-b83b-4afd-b414-1be5bd41b39e

pdanpdan:
This seems to be related to `visibility: visible` set in https://github.com/saadeghi/daisyui/blob/1569957f38f659e9eba87a7e56a382f19c499470/packages/daisyui/src/components/collapse.css#L2

@saadeghi why is the first rule there? I cannot see a reason, and git blame does not help


https://play.tailwindcss.com/skTqwxHMpY?file=css

saadeghi:
@pdanpdan  
Class name conflict: https://tailwindcss.com/docs/visibility  
`collapse` class name was added to Tailwind CSS after daisyUI introduced the `collapse` component class name.  
Since Tailwind's `collapse` is an uncommon class name to use and is unrelated to daisyUI's `collapse` this selector avoids the conflict.

pdanpdan:
Then maybe it should be changed to `visibility: unset`?
Because `visibility: visible` produces some funny side effects when in a modal (the funny thing is that it makes an input inside the collapse autofocus, and this leads to browser scrolling it into view, which breaks the animation)

pdanpdan:
[...]

Issue #3440, opened 2025-01-31
bug: V5 Dialog does not put focus on first focusable element


mathieulesniak:
### What version of daisyUI are you using?

v5 beta 6

### Which browsers are you seeing the problem on?

Chrome

### Reproduction URL

https://codepen.io/mlesniak/pen/MYgxPxr

### Describe your issue

**Context:** 
When opening a `dialog` html element, focus is automatically set to first focusable element in the dialog.
In DaisyUI v5 beta 6, the first focusable element does not take focus once the modal is opened (JS way)
In addition, trying to focus an `input` with JS does not seems to work either.

In v4, focus was ok when opening a modal : https://codepen.io/mlesniak/pen/ogvVaOR (whereas text field does not show blinking caret)

**Steps to reproduce:** 
* open the according codepen (v4 or v5)
* click on button open modal in daisy UI
* try to type something. Typed text should appear in modal input

saadeghi:
fixed in `5.0.0-beta.7`

ymmt07039:
It seems that this issue has reoccurred in 5.0.0-beta.9. It was working fine in beta.8.

Conni2461:
Also experiencing this issue in 5.0.0. Looking at the commits i think https://github.com/saadeghi/daisyui/commit/131eabd9516ca0e76fe7e9b7a7e10022582146ef reintroduced the issue. 

The codepen for reproducing also still seems valid, you just have to replace the version, none the less should i open a new issue?

delasource:
This can be easily seen on the demo of https://nexus.daisyui.com/dashboards/ecommerce - just click the search button on the top (it should automatically set focus to the input)

saadeghi:
I'm working on it

delasource:
seems to be fixed! 👍  thank you

neopointer:
@saadeghi I also have the same issue. I've tried in the latest version and it doesn't work.

Even calling `element.focus()` on the input field inside of the modal does not put the focus on the element.

Is there something especial that needs to be done?

saadeghi:
@neopointer 
> Is there something especial that needs to be done?

No.  
This is fixed already. If it's not working for you, please share a minimal reproduction link so I can help.

neopointer:
@saadeghi  here's the reproducer with the latest version: https://codepen.io/neopointer/pen/OPPjvGq

Here's another version _without daisy ui_: https://codepen.io/neopointer/pen/XJJaqrM

[...]

Issue #2223, opened 2023-08-06
bug: <dialog> element items are focusable, even when <dialog> element is closed


viktorb1:
### What version of daisyUI are you using?

3.5.1

### Describe your issue

When you use a dialog modal such as https://daisyui.com/components/modal/#dialog-modal, everything is still tabbable even when the modal is not open. The problem seems to be caused by display: grid preventing the default behavior of the <dialog> element from being set to display: none.

### What browsers are you seeing the problem on?

Chrome

### Reproduction URL

[Your website demos, for example](https://daisyui.com/components/modal/#dialog-modal)

saadeghi:
Thank you for reporting.  
Unfortunately removing the `grid` (and adding it only when modal is open) disables the open animation.  
I don't know if there's a way to fix that yet, but I don't want fixing one issue to create another.

viktorb1:
In my project, I dynamically add/remove the 'modal' class when the <dialog> element opens/closes and it works as expected. Decent workaround for the time being if anyone else runs into this problem.

dogagenc:
@saadeghi This is being a problem both for general keyboard usage and accessibility. Is there a reason you don't want to use `visibility: hidden` and `visibility: visible` for close/open states? - which seems it's being used in `unstyled` components - 

It shouldn't affect animation.

velut:
A quick solution (sacrificing the open animation) to restore the default `display:none` behavior seems to be using `open:modal` instead of `modal` as the dialog class.

```html
<dialog id="my_modal_1" class="open:modal">
  <!-- Change the class here   ^^^^^^^^^^ -->
</dialog>
```

cknight:
> In my project, I dynamically add/remove the 'modal' class when the element opens/closes and it works as expected. Decent workaround for the time being if anyone else runs into this problem.

I've just run into this and your workaround is good, but doesn't work if the user presses escape to close the dialog.  In this case, the dialog closes, but still contains the `modal` class.  Did you find a workaround for that?

viktorb1:
@cknight It's been a while since I've had this issue, but what is stopping you from writing a keypress event handler in javascript?

oyvindlille:
Any progress on this? This is in conflict of multiple wcag critera like 2.4.3.
My fix was to use open:modal, but suggest this to be fixed on the class .modal or at least be mentioned in the documentation.

Atmos4:
Achieving this in pure CSS is from my experience very hard. 
[...]

PR #4161, opened 2025-10-10
fix(modal): make sure elements inside modal are hidden when not open

pdanpdan:
The previous fix c1e2f0ccfbd5c324f90bbe84168b6261fb56148a solved all the animation/focusing problems when opening #3440, #3835 But as a side effect it broke #2223

Combined with #4160 these changes fix also that problem It also unifies the styles for opened modal

https://play.tailwindcss.com/Jjq1xK4zjy?file=css


saadeghi:
Amazing 👏

PR #4171, opened 2025-10-11
feat(countdown): cycle last digit between 0-100, allow setting the min number of digits, prevent RTL bug

pdanpdan:
To force a minimum number of digits (use `--value-pad: 2` or `--value-pad: 3`)
Updates the docs to show 2 digits in clock-style examples

Example: https://play.tailwindcss.com/m8wgZ4ATz3?file=css


saadeghi:
Nice 👍  
I think `--digits` would be a better name instead of `--value-pad`:

What's the value? `--value: 34`
How many digits? `--digits: 3`
=> `034`

pdanpdan:
I first thought of pad, then of digits, then of pad again :(
Can you easily rename to digits or should I?

saadeghi:
I'll do 👍

PR #4175, opened 2025-10-11
fix(countdown): fix safari miss-alignment when zoom does not result in an integer for line height

### LINKED ISSUE #3872: bug: Countdown unaligned at 85% zoom in safari Sequoia 15.4.1

https://daisyui.com/components/countdown/?lang=en

5.0.35

Safari

### Describe your issue

When zooming to 85% on safari, the countdown becomes unaligned. Half of the current value is hidden under the hidden overflow and half of the next value is showing

Image

Also occurring on daisyui@4.12.10. This did not occur for me on Firefox@136.0 on Ubuntu 20.04.6.

pdanpdan:
safari rounds 1em when used in height but not when used in top

close #3872

### DIFF
--- a/packages/daisyui/src/components/countdown.css
+++ b/packages/daisyui/src/components/countdown.css
@@ -1,7 +1,8 @@
 .countdown {
   @apply inline-flex;
   &.countdown {
-    line-height: 1em;
+    --em: round(up, 1em, 1px);
+    line-height: var(--em);
   }
 
   & > * {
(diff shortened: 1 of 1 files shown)


pdanpdan:
Ok, sorry. Safari is really something special. I write height: 16px and it calculates 15.99px
I'll close this.

saadeghi:
Thanks for trying 🫡  

> Safari is really something special.


a8szoj

PR #4189, opened 2025-10-16
feat: add modifier validator-hint-hidden to hide the hint when not visible

### LINKED ISSUE #4188: bug: validator-hint hidden not work any more, v5.0.33 works

_

5.3.3

_No response_

### Describe your issue

If you prefer to hide the hint when it's not visible, add hidden class to validator-hint.
It will take no space when it's hidden, and if input is invalid, it will appear and take space.
but it not works any more in the new version

pdanpdan:
using tw hidden does not work anymore because it has higher priority
example: https://play.tailwindcss.com/0DpbDoVzqO

I could not update the translation for `fa.json`

close #4188

### DIFF
--- a/packages/daisyui/src/components/validator.css
+++ b/packages/daisyui/src/components/validator.css
@@ -22,7 +22,7 @@
         --input-color: var(--color-error);
       }
       & ~ .validator-hint {
-        @apply text-error visible block;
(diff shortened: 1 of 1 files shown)


saadeghi:
I think we should fix the priority instead.

PR #4199, opened 2025-10-20
feat: improve collapse to allow searching in content for details variant

pdanpdan:
ref #4194
demo:  https://play.tailwindcss.com/e1xWGNi26S?file=css

### DIFF
--- a/packages/daisyui/src/components/collapse.css
+++ b/packages/daisyui/src/components/collapse.css
@@ -42,6 +42,9 @@
       > :where(input:is([type="checkbox"], [type="radio"]):checked ~ .collapse-content) {
       content-visibility: visible;
       min-height: fit-content;
+      @supports not (content-visibility: visible) {
+        visibility: visible;
+      }
     }
     &:focus-visible,
     &:has(> input:is([type="checkbox"], [type="radio"]):focus-visible),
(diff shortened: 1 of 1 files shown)


pdanpdan:
The only problem this might generate is that [caniuse](https://caniuse.com/css-content-visibility) only declares ios safari from 18.0 supporting it (2%).

Please let me know it we need to cover it, so that I change the styles

pdanpdan:
I have added fallback to visibility for ios < 18
Do you happen to have how to test on a device with such a browser?

saadeghi:
Nice   
I will test on iOS simulator.

pdanpdan:
I do not expect find inside collapse to work there, but collapse should work as before
I mean focus and checkbox variants should be animated, and I believe details variant should not be animated (I think this is how it works now)

saadeghi:
I tested on iOS 26 and iOS 17.5  
Everything's good. but in general, iOS does not search text inside `<details>`.

9 search results on desktop, 3 on mobile:

https://github.com/user-attachments/assets/6de327b4-fcb7-4857-8527-44413c2ab093

pdanpdan:
Yes, working without search is expected on iOS - that is what they can do :)

saadeghi:
Thanks for the PR @pdanpdan

anthony-robin:
Thank you both for working on this feature so quickly :pray:

pdanpdan:
I'll open another PR to document the feature in docs

saadeghi:
@pdanpdan thanks. It can be just one line, about how browser text search can find the text in `<details>`.  
No need to over explain or talking about hidden until find...

PR #4203, opened 2025-10-21
Use Zed instead of Cursor in Zed LLM setup docs

Flo0807:
- Use Zed instead of Cursor in Zed LLM setup docs


saadeghi:
Thank you

PR #4204, opened 2025-10-21
Update Phoenix install example

Flo0807:
The command `mix phx.new ./ --no-ecto` results into:

> ** (Mix) Application name must start with a letter and have only lowercase letters, numbers and underscore, got: "Projects". The application name is inferred from the path, if you'd like to explicitly name the application then use the `--app APP` option.

because my directory is named "Projects" (note the uppercase letter).

This PR updates the command to use an explicit name (`my_app`).


saadeghi:
Thanks for the PR, but I think the error is clear enough.  
The command is supposed to `Create a new Phoenix project in the current directory.`  
And your **current** directory name is not accepted because of the reason you mentioned.  

We're not creating a new directory for the project forcefully because people may use different file structure for their projects. It could be an existing project, it could be a monorepo with existing files, etc.

PR #4214, opened 2025-10-23
fix: prevent tab style bleeding to children tabs

pdanpdan:
This is a quick fix for the tabs problem (#4200).

It applies the missing direct children relationship that was applied already for lift

Please let me know it you can think of a situation where this will break something that now is working (docs lgtm with changes).

Example: https://play.tailwindcss.com/jyeC1BSBHZ?file=css

### DIFF
--- a/packages/daisyui/src/components/tab.css
+++ b/packages/daisyui/src/components/tab.css
@@ -21,9 +21,12 @@
     --tab-radius-ee: 0;
     --tab-order: 0;
     --tab-radius-min: calc(0.75rem - var(--border));
-    --tab-rm: min(var(--radius-field), var(--tab-radius-min));
-    --tab-grad: calc(69% - var(--border));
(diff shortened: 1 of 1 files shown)


packages/daisyui/src/components/tab.css
```diff
@@ -494,15 +494,15 @@
       using --tabs-box-radius CSS variable to avoid Nuxt build issue
       https://github.com/saadeghi/daisyui/issues/3724
     */
-    --tabs-box-radius: calc(var(--radius-field) + var(--radius-field) + var(--radius-field));
+    --tabs-box-radius: calc(3 * var(--radius-field));
```
pdanpdan: Was there a specific reason why this was written as a sum?
I understand why it was extracted in a variable

saadeghi: This line evolved to this slowly. 
Originally it was in a min() in a another calc() and some PostCSS build tools were merging them all, resulting wrong values.  
https://github.com/saadeghi/daisyui/commit/ac358b7ebf7747553cf5fb53ddb718cfb585abba

I moved it to the variable as it was, But now it can indeed be written simpler 👍

saadeghi:
Thank you 🙏

PR #4216, opened 2025-10-24
fix: allow outline on all activators for dropdown, not only summary

pdanpdan:
All direct children of dropdown before a dropdown-content should show focus indicators

example not working: https://play.tailwindcss.com/AoGt1QaprX


saadeghi:
Changed it to `& > *:not(:has(~ [class*="dropdown-content"])):focus {` so it works with prefix too.

PR #4223, opened 2025-10-29
docs: labels should be associated to inputs

tkrotoff:
Labels and inputs should be associated in the documentation:

```HTML
<label for="myInput">
<input id="myInput">
```

Otherwise the form is not accessible.

Here VoiceOver output:

- Before:

image

<br><br>

- After:

image

### DIFF
--- a/packages/docs/src/routes/(routes)/components/fieldset/+page.md
+++ b/packages/docs/src/routes/(routes)/components/fieldset/+page.md
@@ -52,32 +52,30 @@ classnames:
 ```
 
 ### ~Fieldset with multiple inputs
+#### Add proper id and for attributes for accessibility. The id must be unique and depends on your context.
 
 <fieldset class="fieldset w-xs bg-base-200 border border-base-300 p-4 rounded-box">
   <legend class="fieldset-legend">Page details</legend>
- 
-  <label for="title" class="label">Title</label>
-  <input type="text" id="title" class="input" placeholder="My awesome page" />
-
-  <label for="slug" class="label">Slug</label>
-  <input type="text" id="slug" class="input" placeholder="my-awesome-page" />
-
-  <label for="author" class="label">Author</label>
-  <input type="text" id="author" class="input" placeholder="Name" />
+  <label class="label">Title</label>
+  <input type="text" class="input" placeholder="My awesome page" />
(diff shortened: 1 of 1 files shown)


saadeghi:
What happens if I copy this in my page, and I already have an element with `id="title"`?
```html
<input type="text" id="title" class="input" placeholder="My awesome page" />
```

tkrotoff:
> What happens if I already have an element with id="title"?

Good point, we can do like [Bootstrap examples](https://getbootstrap.com/docs/5.3/forms/form-control/#example): "exampleFormControlInput1"

```HTML
<label for="exampleFormControlInput1" class="form-label">Email address</label>
<input type="email" class="form-control" id="exampleFormControlInput1" placeholder="name@example.com">
```

saadeghi:
What happens if there are multiple copies of same item?
```html
<label for="exampleFormControlInput1" class="form-label">Email address</label>
<input type="email" class="form-control" id="exampleFormControlInput1" placeholder="name@example.com">

<label for="exampleFormControlInput1" class="form-label">Email address</label>
<input type="email" class="form-control" id="exampleFormControlInput1" placeholder="name@example.com">

<label for="exampleFormControlInput1" class="form-label">Email address</label>
<input type="email" class="form-control" id="exampleFormControlInput1" placeholder="name@example.com">
```

For example a login form component that is imported in a navbar dropdown and in a footer.

We should not decide for the IDs because duplicate ID will lead to invalid HTML and literal app disfunction, and you wouldn't even know. Linters can't even help with that because they check the source code, not the result.  
Meanwhile, lack of ID will give you an accessibility warning (as expected) and guides you to adding an ID.

tkrotoff:
If you feel it's risky to add ids in examples, then the doc should explain how to make the code snippets accessible.

Something like _"Warning: this code snippet is not accessible; add [for and id attributes to label and input](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/for) to make it [accessible](https://en.wikipedia.org/wiki/Web_accessibility)"_

tkrotoff:
I've opened a discussion on examples being accessible out of the box: https://github.com/saadeghi/daisyui/discussions/4236

PR #4232, opened 2025-11-01
feat: keep colorState in palette instead of slider and allow lossless change of color mode

pdanpdan:
It's an initial attempt. It should work as expected, but I'm not sure about performance.


pdanpdan:
Now it should be good

saadeghi:
Thank you @pdanpdan

PR #4234, opened 2025-11-02
feat: use auto height drawer when drawer-open (update docs to explain)

pdanpdan:
ref #4233

- the changes if the navbar example are in order to hide the drawer open as overlay when screen size increases

### DIFF
--- a/packages/daisyui/src/components/drawer.css
+++ b/packages/daisyui/src/components/drawer.css
@@ -137,7 +137,7 @@
 
   @layer daisyui.modifier {
     & > .drawer-side {
-      @apply overflow-y-auto;
+      @apply h-auto overflow-y-auto;
     }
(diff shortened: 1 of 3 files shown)


saadeghi:
Thanks for the PR, but we can't do it now. It's a breaking change, it will change a behavior on all the projects using daisyUI right now and it requires everyone to manually add class name to get the same behavior.  

Can't merge this 😅

pdanpdan:
Then I would add it as a modifier. It's pretty hard to get rid of the height otherwise

saadeghi:
Adding a `md:h-auto` works, no?  
https://play.tailwindcss.com/FDHEYkh224?size=794x542

pdanpdan:
:(
Of course it works

PR #4235, opened 2025-11-02
fix: height of tab-content for tabs-box (compensate for padding)

### LINKED ISSUE #4227: bug: radio tabs-box + tab content don't work well with height on .tabs

https://play.tailwindcss.com/2UArsMGvWZ

5.3.10

chrome 141.0.7390.123

### Describe your issue

Image

When .tabs has a height, its bottom padding gets covered by the .tab-content.

pdanpdan:
- .tab-content is hidden by default so I think it doesn't make sense to set the height only when visible

close #4227

example: https://play.tailwindcss.com/vGGKdhWbYf


saadeghi:
Thank you 🌼

PR #4246, opened 2025-11-06
fix(format): use the format command 

WuMingDao:
Use the format command


saadeghi:
Thanks for the PR, but I can't verify these changes 😅  
I guess it would be easier if run it in another commit.

PR #4255, opened 2025-11-06
fix: menu disabled visual and docs closes: #4251

### LINKED ISSUE #4251: bug: Disabled menu items still work with keyboard

https://play.tailwindcss.com/YZF6oUUS3p

v5.4.5

Chrome

## The Problem

  Disabled menu items can't be clicked with a mouse (good), but you can still Tab to them and activate them with Enter/Space (bad).

  ## How to Reproduce

  ```html
  <ul class="menu bg-base-200 w-56 rounded-box">
    <li><a href="#enabled">Enabled Item</a></li>
    <li class="menu-disabled"><a href="#disabled">Disabled Item</a></li>
    <li><a href="#enabled2">Enabled Item 2</a></li>
  </ul>

  1. Press Tab to navigate through the menu
  2. The disabled item gets focus
  3. Press Enter or Space - it activates!

sulimanbenhalim:
## Problem

Fixes #4251

Disabled menu items (with `menu-disabled` class) only have visual opacity styling but still show focus states when tabbing into them, which is confusing

Also, the CSS cannot prevent keyboard focus, only HTML attributes like `inert` or `tabindex="-1"` can do that, and docs do not mention this

## Fix

1. Added CSS rules to completely remove all interactive visual feedback (hover, focus, active states) from disabled menu items
2. Added documentation note explaining that `menu-disabled` provides visual styling only and users should add `inert` or `tabindex="-1"` for full keyboard accessibility

## Testing

Tested in playground - disabled items no longer show any hover/focus/active visual feedback

### DIFF
--- a/packages/daisyui/src/components/menu.css
+++ b/packages/daisyui/src/components/menu.css
@@ -111,7 +111,22 @@
       }
 
       &.menu-disabled {
-        @apply text-base-content/20 pointer-events-none;
+        @apply text-base-content/20 pointer-events-none cursor-not-allowed;
+
+        & > *:not(ul) {
+          @apply cursor-not-allowed;
+
+          &:hover,
+          &:focus,
+          &:focus-visible,
+          &:active,
+          &.menu-active {
+            @apply outline-hidden bg-transparent;
+            box-shadow: none;
+            color: inherit;
+            cursor: not-allowed;
+          }
+        }
       }
     }
     .dropdown:focus-within {
--- a/packages/docs/src/routes/(routes)/components/menu/+page.md
+++ b/packages/docs/src/routes/(routes)/components/menu/+page.md
@@ -434,6 +434,10 @@ classnames:
 </ul>
 ```
 
+> :INFO:
+>
+> `menu-disabled` provides visual styling only. To prevent keyboard focus, add `inert` attribute or `tabindex="-1"` to the disabled item.
+
 
 ### ~Menu with icons
 <ul class="menu bg-base-200 w-56 rounded-box">


pdanpdan:
Please rebase the PR branch on current master from main repo

And the CSS part only hides the problem - a11y tools will still announce the menu item as active.

I would go simply with fixing the docs example - `a` is always enabled, so the example should use `div` or better `span`

sulimanbenhalim:
@pdanpdan The branch is already rebased on the current master

sulimanbenhalim:
@pdanpdan With respect, I disagree with your assessment.

The CSS changes provide **essential visual feedback improvements**, disabled items should not show hover/focus/active states! This is a UX improvement regardless of a11y concerns

The documentation I added **explicitly addresses the a11y concern** by instructing users to add `inert` or `tabindex="-1"` attributes, this is the correct semantic HTML solution that a11y tools will respect

> Please rebase the PR branch on current master from main repo
> 
> And the CSS part only hides the problem - a11y tools will still announce the menu item as active.
> 
> I would go simply with fixing the docs example - `a` is always enabled, so the example should use `div` or better `span`

sulimanbenhalim:
Changing the docs example to use `<div>` or `<span>` instead of `<a>` doesn't solve anything, users already legitimately need to use `<a>` and `<button>` elements in menus, so the proper solution is documenting the
correct HTML attributes, which this PR does

> Please rebase the PR branch on current master from main repo
> 
> And the CSS part only hides the problem - a11y tools will still announce the menu item as active.
> 
> I would go simply with fixing the docs example - `a` is always enabled, so the example should use `div` or better `span`

pdanpdan:
While inert is a good option, `tabindex="-1"` is not.
https://css-tricks.com/how-to-disable-links/#aa-just-dont-do-it

pdanpdan:
Can you please also fix the examples?
But I'm still of the opinion that adding `10 lines of CSS, 15 lines of JavaScript` to almost fix something that should not be there is not the right thing to do.

saadeghi:
We don't use `!important` in daisyUI.

sulimanbenhalim:
@pdanpdan You're contradicting W3C standards and every major design system, Here's the facts:

W3C explicitly supports disabled menu items

From the official W3C ARIA Authoring Practices Guide (w3.org/WAI/ARIA/apg/patterns/menubar/):
- "Disabled menu items are focusable but cannot be activated"
- "When a menu item is disabled, aria-disabled is set to true"

So your claim that disabled menu items "should not be there" contradicts the W3C specification

> Please rebase the PR branch on current master from main repo
> 
> And the CSS part only hides the problem - a11y tools will still announce the menu item as active.
> 
> I would go simply with fixing the docs example - `a` is always enabled, so the example should use `div` or better `span`

sulimanbenhalim:
W3C's own examples use <a> tags

From w3.org/WAI/ARIA/apg/patterns/menu-button/examples/menu-button-links/:
- "The menuitem role is on the HTML ```<a>``` element contained by each ```<li>``` element so link behaviors are available when focus is set on the menu item"

The W3C uses anchor tags in their examples. Your suggestion to use ```<div>``` or ```<span>``` goes against their implementation.

> I would go simply with fixing the docs example - `a` is always enabled, so the example should use `div` or better `span`

saadeghi:
Explained here https://github.com/saadeghi/daisyui/issues/4251#issuecomment-3499763245

saadeghi:
> The W3C uses anchor tags in their examples. Your suggestion to use `<div>` or `<span>` goes against their implementation.
> 
> >

Doesn't apply here because in our case if we don't want it to be used (clicked or navigated), it's not an actual link. it's a fake link. Looks like a link, but it's not.

sulimanbenhalim:
You claim tabindex="-1" shouldn't be used

Bootstrap's official documentation explicitly recommends it:
- "Include a tabindex="-1" attribute on these links to prevent them from receiving keyboard focus"

So does Headless UI, Radix UI, and Chakra UI. All major design systems use tabindex="-1" for disabled menu items.

Your linked article is from 2017 and talks about generic link disabling, not menu components. It doesn't apply here

> While inert is a good option, `tabindex="-1"` is not. https://css-tricks.com/how-to-disable-links/#aa-just-dont-do-it

sulimanbenhalim:
@saadeghi With respect, these aren't "fake links", they're real navigation items that are temporarily disabled, so we can call this is a standard UI pattern

for example if we look at any application menu:
- File -> Save (disabled when nothing changed)
- Edit -> Paste (disabled when clipboard empty)

These are real links that become available when conditions change so they're not fake they're just conditional

> > The W3C uses anchor tags in their examples. Your suggestion to use `<div>` or `<span>` goes against their implementation.
> > >
> 
> Doesn't apply here because in our case if we don't want it to be used (clicked or navigated), it's not an actual link. it's a fake link. Looks like a link, but it's not.

sulimanbenhalim:
@saadeghi what happens if we use ```<div>``` or ```<span>```

we will loses semantic HTML (hurts SEO and accessibility)
we will break existing user code (breaking change)
and we will go against W3C examples and every major design system

saadeghi:
I found this on W3C:  
https://w3c.github.io/html-aria/#example-communicate-a-disabled-link-with-aria

> In the following example, a hyperlink needs to be communicated as being in the disabled state. HTML does not allow for the use of the disabled attribute on a hyperlink, and using aria-disabled=true would communicate the hyperlink as being disabled to assistive technologies, but would not actually disable the element. The most effective way to both communicate and actually disable a hyperlink would be to remove the href from the [a](https://html.spec.whatwg.org/multipage/text-level-semantics.html#the-a-element) element, creating a placeholder. Then, ARIA can be applied to this placeholder link to communicate the element's intended role and state.
> ```html
>   <a role=link aria-disabled=true>...</a>
>  ```

And already works well with daisyUI:  
https://play.tailwindcss.com/XMqdCtuxXl

The docs can be changed to `<a role="link" aria-disabled="true">` instead of `<a>`

PR #4256, opened 2025-11-06
fix(tooltip): reorganize selectors for active tooltip and fix empty tooltip-content on focus

### LINKED ISSUE #4254: bug: Empty tooltips show when using Tab

https://play.tailwindcss.com/PeCESEOPAG

V5.4.5

_No response_

## The Problem

Tooltips with `data-tip=""` (empty) 

correctly don't show on hover, but they DO show as empty boxes when you Tab to focus the element

## How to Reproduce

```html
<div class="tooltip" data-tip="">
  <button class="btn">Focus me with Tab</button>
</div>

<div class="tooltip" data-tip="Real tooltip">
  <button class="btn">This one works fine</button>
</div>
```

Image

1. Click on the page
2. Press Tab to focus the first button
3. Empty tooltip box appears (shouldn't)

pdanpdan:
close #4254

reorganize the selectors so it is more clear what is displayed

### DIFF
--- a/packages/daisyui/src/components/tooltip.css
+++ b/packages/daisyui/src/components/tooltip.css
@@ -18,16 +18,6 @@
       content: var(--tw-content);
     }
 
-    @media (prefers-reduced-motion: no-preference) {
-      & > .tooltip-content,
-      &[data-tip]:before,
-      &:after {
-        transition:
-          opacity 0.2s cubic-bezier(0.4, 0, 0.2, 1) 75ms,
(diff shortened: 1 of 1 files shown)


saadeghi:
I think this is breaking tooltip tail

pdanpdan:
Let me check, please

No, I just checked again both in docs and in the example https://play.tailwindcss.com/GNQnQO67Nl?file=css

The image is from the docs?

PR #4257, opened 2025-11-06
fix: calendar disabled hover closes: #4253 

### LINKED ISSUE #4253: bug: Calendar disabled dates show hover effect

https://daisyui.com/components/calendar/

v5.4.5

_No response_

## The Problem

When using react-day-picker, disabled dates still show the hover background color
Makes them look clickable when they're not

## How to Reproduce

```jsx
import { DayPicker } from "react-day-picker"

<DayPicker
  mode="single"
  disabled={{ from: new Date(2000, 0, 1), to: new Date() }}
/>
```

Image

Hover over any disabled (past) date - it shows the hover background!

sulimanbenhalim:
## Problem

Fixes #4253 

Disabled calendar dates in react-day-picker show hover background color, making them appear interactive when they cannot be clicked which creates confusing UX

## Fix

Added CSS rule to remove hover background from disabled date buttons (`:disabled:hover` and `[aria-disabled="true"]:hover`) and set cursor to `not-allowed`

## Testing

Tested in playground at disabled dates no longer show hover background color

### DIFF
--- a/packages/daisyui/src/components/calendar.css
+++ b/packages/daisyui/src/components/calendar.css
@@ -110,6 +110,11 @@
       &:hover {
         background-color: var(--color-base-200);
       }
+      &:disabled:hover,
+      &[aria-disabled="true"]:hover {
+        background-color: transparent;
+        cursor: not-allowed;
+      }
     }
     .rdp-caption_label {
       z-index: 1;


pdanpdan:
Please, please, fix the content of the PR

sulimanbenhalim:
> Please, please, fix the content of the PR

@pdanpdan Give me a minute to fix things before spamming comments

saadeghi:
Have some tea ☕️☕️🫖 @sulimanbenhalim @pdanpdan

pdanpdan:
I would bo with `:not([disabled])` on the rule above. The days are buttons with `disabled` set or not.
So the cursor is natively set, and now it is a double rule to set and unset the background

saadeghi:
@pdanpdan  Yes that would be cleaner, but I think it would require adding not in all other states such as `rdp-selected`, `rdp-range_start`, `rdp-range_middle` etc (I didn't test yet)

PR #4286, opened 2025-11-14
feat(docs): work on theme picker

pdanpdan:
- fix and optimize inputValue processing in ColorPalette
  - svelte reactivity is not very funny
  - the bindable props changes are called right when the value changes (not batched), and we have to be very carefull with the order different state values are changed

- use 3 decimals for `c` and `h` in `oklch` format to match tw palette format/precission

- use a single handleKeydown listener in ColorPalette

- use colorName as key to mark colors as selected (seems more natural, as color name is already calculated)

- make sure we can distinguish neutral-50 instead of zinc-50 (in tailwind palette neutral-50 and zinc-50 are identical)

- avoid bind:value in ThemeCSSModal because we already have a listener on it


saadeghi:
Thanks for the PR @pdanpdan 

The value of `neutral-50` and `zinc-50` is indeed the same.  
But isn't it a weird UX that people cannot choose `zinc-50`.  
It feels like that color is broken if we click it and another cell gets selected.  
I think either both should be shown as selected (existing behavior at https://daisyui.com/theme-generator) or the chosen one must be selected.

pdanpdan:
Let me think about this.

pdanpdan:
I agree that clicking on `zinc-50` and not seeing it as active is somehow strange.

At the same time clicking on `neutral-50` and seeing that the name of the selected color is `zinc-50` is also strange (current behavior - and I would expect almost no one will think of the color as zinc but as neutral) :).

I have thought of a workaround - add a text-only change to one of the colors:
`"zinc-50": "oklch(98% 0 0)"` => `"zinc-50": "oklch(98% 0.000 0.000)"`.
Using 3 decimals will also fit the way `zinc-xxx` colors are defined.

I will revert the changes that were replacing `zinc-50` with `neutral-50`.

After the changes, selecting a color from palette will display that color as selected (and the right color name in the edit field).

@saadeghi The changes are implemented.

saadeghi:
What if we track the Tailwind color name as well?  
https://github.com/saadeghi/daisyui/pull/4286/commits/bdb1f8d726b3f2af99fd8d7cdc83128ea2674ae6

This way, it shows the correct color name, regardless of the identical color value, and there would be no need to manually edit the color values to make them different from each other.  
I think this would be cleaner.

pdanpdan:
The change does not solve the problem of color name identification after theme is loaded, because we only have the value. 
I mean when we use the light theme, base-200 will always be named zinc-50. 

:) and for all the other colors the name - value relation is 1 to 1

saadeghi:
That's not an issue. I mean that's expected, when CSS has `oklch(98% 0 0)` it doesn't matter if the color picker translates it to zinc or neutral…
But it matters if the user of color picker (who is not aware that the value of both color names are the same) see the color name they chose.

pdanpdan:
Ok, then it looks good to me.

saadeghi:
> :) and for all the other colors the name - value relation is 1 to 1

Yes, but I don't think it's a good idea to modify the data source (color values). Tailwind might add new color names in a future release (imagine a new `*-990` darkness) and some other color names may have identical value.  
It wouldn't be future proof if modifying the color value in the data source is required (things like `0`->`0.000`)

saadeghi:
Thank you for the PR @pdanpdan 🌼

PR #4290, opened 2025-11-15
fix(button): move color classes in layer l3 (component)

### LINKED ISSUE #4288: bug: buttons with light text in disabled state are unreadable

https://play.tailwindcss.com/I3gKDNHK5Z

5.5.3

Chrome

### Describe your issue

button variants that have light text become unreadable when disabled with the `btn-disabled` class. This class doesn't override the text of the button regardless of its position when specifying the class names. This is especially relevant in light themes

Image

pdanpdan:
- only set tw-content if aria-label is present to keep tw-content from filter reset button (does not check if aria-label is empty - but I would say it's a user decision)

example: https://play.tailwindcss.com/uM3Yd12gIl
please let me know if some combination does not look as expected

close #4288

### DIFF
--- a/packages/daisyui/src/components/button.css
+++ b/packages/daisyui/src/components/button.css
@@ -110,56 +110,56 @@
 }
 
 .btn-primary {
-  @layer daisyui.l1.l2 {
+  @layer daisyui.l1.l2.l3 {
     --btn-color: var(--color-primary);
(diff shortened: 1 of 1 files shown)


saadeghi:
Looks great. Thanks @pdanpdan

PR #4292, opened 2025-11-15
cleanup a few instances of calc in calc

pdanpdan:
- they are not a problem but it would be cleaner without them
- while cleaning up I discovered that in tabs we are adding and substracting the same thing :)


saadeghi:
Thank you @pdanpdan

PR #4296, opened 2025-11-16
fix(badge): calculate inline padding based on --size-selector

### LINKED ISSUE #4293: bug: Empty badges lose their 1:1 ratio when theme --size-selector is modified

https://play.tailwindcss.com/GjYkrVib7m

5.5.5

Chrome

### Describe your issue

Image

After modifying the default `--size-selector` at the theme level, which the theme editor allows, empty badges lose their square ratio. The issue is that the `padding-inline` attribute uses a hardcoded value `0.25rem` matching the default selector value, which makes the issue not obvious since all official themes use this value in their `--size-selector`. The suggested fix is to use the --size-selector variable instead of the hardcoded `0.25rem` to adjust the horizontal padding.

I suggest replacing the following expression in all sizes: 
```css
padding-inline: calc(0.25rem * 3 - var(--border));
```
with: 
```css
padding-inline: calc(var(--size-selector) * 3 - var(--border));
```

pdanpdan:
- keeps the badge aspect 1:1 when empty
- simplifies the CSS

close #4293

It looks like button, kbd, and tab also have size based on multiples of --size-fiels or --size-selector and paddings based on multiples of 0.25rem.

Should we also convert them to use inline padding based on --size-fiels or --size-selector?

### DIFF
--- a/packages/daisyui/src/components/kbd.css
+++ b/packages/daisyui/src/components/kbd.css
@@ -2,8 +2,7 @@
   box-shadow: none;
   @layer daisyui.l1.l2.l3 {
     @apply bg-base-200 rounded-field inline-flex items-center justify-center align-middle;
-    padding-left: 0.5em;
-    padding-right: 0.5em;
+    padding-inline: 0.5em;
     border: var(--border) solid color-mix(in srgb, var(--color-base-content) 20%, #0000);
     border-bottom: calc(var(--border) + 1px) solid
       color-mix(in srgb, var(--color-base-content) 20%, #0000);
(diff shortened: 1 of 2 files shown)


simoami (approved):

simoami:
>Should we also convert them to use inline padding based on --size-fiels or --size-selector?

@pdanpdan I think this is a problem with components that have an expected aspect ratio that ends up being broken due to theme size adjustments. I don't think the other components you mentioned have a case where aspect ratio needs to be locked.

PR #4297, opened 2025-11-17
Update Tailwind CSS installation command

Martini991:
I updated the installation command to include the -D flag because TailwindCSS, its Vite plugin, and DaisyUI (as I understood) run only during the build process and are not required at runtime.


saadeghi:
Thank you

PR #4302, opened 2025-11-19
docs(validator): explain that validator-hint must have only on previous sibling validator

### LINKED ISSUE #4300: bug: validator-hint should only consider previous validator

https://play.tailwindcss.com/xXS2s1ZDiQ

v5.5.5

All browsers

### Describe your issue

When having multiple inputs in a form, each with the validator class, if only one is in an invalid state, the hints of all following inputs are also displayed, instead of only the hint of the invalid input.

Image

In the above example, the Count hint should not be shown, since the Count input is in a valid state, as can be seen by the green border of the validator class.

pdanpdan:
close #4300


saadeghi:
Thank you

PR #4314, opened 2025-11-27
perf: optimize async operations and function chaining

swarnim02:
- Remove unnecessary async/await in extractClasses map callback
- Chain CSS transformations in cleanCss for better performance
- Minor performance improvements with no functional changes

### DIFF
--- a/packages/daisyui/functions/cleanCss.js
+++ b/packages/daisyui/functions/cleanCss.js
@@ -5,20 +5,11 @@ export const cleanCss = (cssContent) => {
     /var\((--(spacing|width)[\w-]*),\s*((?:[^)(]+|\((?:[^)(]+|\([^)(]*\))*\))*)\)/g
   const spacingVarRegex = /var\(--spacing\)/g
 
-  // Remove empty fallbacks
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks for the PR but what's the benefit of this?

swarnim02:
@saadeghi Hey! Good question - these are small but meaningful optimizations:

**extractClasses.js**: Removed redundant `async` in the map callback. When you're already returning a promise from `processCssFile()`, wrapping it in another async function just adds unnecessary overhead.

**cleanCss.js**: Chained the string operations instead of reassigning the variable multiple times. It's cleaner and slightly more efficient since we're not creating intermediate variables.

Nothing groundbreaking, but these patterns add up when processing lots of files during builds. Plus it makes the code a bit more readable IMO.

Happy to close if you think it's not worth it though!

saadeghi:
Sorry but this PR has no effect on build speed or build performance.

Even if it did, it doesn't effect the distributed library.

Thanks for the effort tho. It would be nice to open a discussion first to talk and agree on a change before putting the effort because I really don't want to discourage contributions 🙏💚

PR #4316, opened 2025-11-28
Fix responsive dropdown rendering under content

Kasparsu:
# What does this commit do
 - Update docs example to include `z-1` so it will render dropdown over content instead of under

### DIFF
--- a/packages/docs/src/routes/(routes)/components/navbar/+page.md
+++ b/packages/docs/src/routes/(routes)/components/navbar/+page.md
@@ -431,7 +431,7 @@ classnames:
       <li>
         <details>
           <summary>Parent</summary>
-          <ul class="p-2">
+          <ul class="p-2 bg-base-100 w-40 z-1">
             <li><a>Submenu 1</a></li>
             <li><a>Submenu 2</a></li>
           </ul>


pdanpdan:
Hello. Can you please explain what does this fix? 
The example shows the drop-down over content now both in small and large screens

Kasparsu:
image

[https://jsfiddle.net/q37pg0re/7/](https://jsfiddle.net/q37pg0re/7/)

example of it not working,

pdanpdan:
I agree with you that in the general case one will probably need to set z-index according to the specific conditions, but the change in the demo code does nothing.

Kasparsu:
> I agree with you that in the general case one will probably need to set z-index according to the specific conditions, but the change in the demo code does nothing.

I'm an educator and I am teaching daisyUI every year and have prolly copied that code hundreds of times and it just bugs the hell out of me that z-1 is missing and I have to add it everytime ;P And since it is added to mobile view dropdown burger menu why not as well there?

pdanpdan:
Ok, but then add it to the source code that is displayed.

Kasparsu:
> Ok, but then add it to the source code that is displayed.

Was that not the correct docs source code?

pdanpdan:
The block just below. The one you modified is what is rendered, the one below is what is shown in the html source tab

Kasparsu:
> The block just below. The one you modified is what is rendered, the one below is what is shown in the html source tab

Oh... didn't notice did in class in hurry but seems you fixed... was going to do it myself anyway cheers and thank you.

PR #4322, opened 2025-11-29
(docs): color picker improve color text representation and make dialog transparent on android

pdanpdan:
- fix ring on selected color in palette on mobiles
- do not add decimals for oklch and hsl if the value is an integer after truncation
- always update hue in oklch (even if chroma is 0)
  - when the color is updated from sliders the hue is not changed when changing chroma
  - when a color is picked from palette the hue slider should reflect selection
  - when text value of the color is changed the hue slider should reflect selection
- make dialog also hide on android when range input is dragged
  - on android the range input loses :active when dragged
  - it needs attaching events for touch and mouse because pointercancel is called on move (even if setPointerCapture is used - probably because svelte re-renders/modifies the control)

### DIFF
--- a/packages/docs/src/components/ColorPalette.svelte
+++ b/packages/docs/src/components/ColorPalette.svelte
@@ -260,11 +260,9 @@
 
   const reNoDecimals = /\.0+$/
   function toFixed(value, decimals) {
-    if (!value) return '0'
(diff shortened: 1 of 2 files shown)


saadeghi:
Thank you @pdanpdan

Issue #3712, opened 2025-03-21
bug: Chat component is not properly aligned in Safari desktop


mtliendo:
### Reproduction URL (Required)

https://daisyui.com

### What version of daisyUI are you using?

v5.0

### Describe your issue

Chrome
<img width="1501" alt="Image" src="https://github.com/user-attachments/assets/2a93e8fe-de0d-4feb-881b-ee6837835710" />

Safari

<img width="1507" alt="Image" src="https://github.com/user-attachments/assets/22da91c2-b06a-4217-bf25-09430d9502dd" />

Code:
```ts
<div
							key={message.id}
							className={`chat ${
								message.sender === currentUser.id ? 'chat-end' : 'chat-start'
							}`}
						>
							{message.sender !== currentUser.id && (
								<div className="chat-image avatar">
									<div className="w-10 rounded-full">
										<img src={stylist.avatar} alt="Stylist avatar" />
									</div>
								</div>
							)}
							<div className="chat-header">
								{message.sender === currentUser.id ? 'You' : stylist.name}
								<time className="text-xs opacity-50 ml-1">
									{formatTime(message.timestamp)}
								</time>
							</div>
							<div
								className={`chat-bubble ${
									message.sender === currentUser.id ? 'chat-bubble-primary' : ''
								}`}
							>
								{message.text}
								{message.imageId && (
									<div className="mt-2">
										<StorageImage
											path={message.imageId}
											alt="Shared image"
											className="rounded-md max-w-full md:max-w-xs max-h-60 object-contain"
										/>
									</div>
								)}
							</div>
						</div>
```

mtliendo:
Gonna close for now while I investigate further. When I load the chat component from DaisyUI's docs in Safari, I don't see the issue. So I must be doing something weird 😅

fowlerwill:
@mtliendo I am able to reproduce on the daisyui doc site by putting more text in one of the chat and squishing the page a bit:

<img width="1255" height="1167" alt="Image" src="https://github.com/user-attachments/assets/3df48c1e-4cb0-4069-809c-ae32ff56fad8" />

pdanpdan:
Please make a play reproduction

fowlerwill:
@pdanpdan https://jsfiddle.net/willfowler/skotjyfn/1/

Safari vs firefox:

[...]

PR #4326, opened 2025-12-02
fix(chat): set explicit grid rows template as min-content to avoid height limitation to height screen on Safari

pdanpdan:
ref: https://github.com/saadeghi/daisyui/issues/3712#issuecomment-3599503502

example: https://play.tailwindcss.com/WRMtKOy4m3?file=css


saadeghi:
Thank you @pdanpdan

PR #4335, opened 2025-12-10
fix(range): do not truncate range selection bar when very wide

### LINKED ISSUE #4334: bug: range input slider displays incorrectly when longer than 3235 pixel

https://play.tailwindcss.com/QFCr6ZiinD?layout=horizontal

tailwind play 4.1.17

Chrome

### Describe your issue

You need a big screen to reproduce this bug !
When the slider is wider than 3235 pixel  (e.g. on a 4K screen) the slider start gets trimmed off when the thumb is at the end. See screenshot from Tailwind play and https://play.tailwindcss.com/QFCr6ZiinD?layout=horizontal

Image

```html
<input type="range" min="0" max="100" value="40" class="range w-full" />
<!-- When this screen is wider than 3235 pixel the slider start gets trimmed off  -->
```

pdanpdan:
- use `100cqw` instead of `100rem`

close #4334


saadeghi:
Thank you @pdanpdan 💚

PR #4337, opened 2025-12-11
fix(list): allow `.list-col-wrap` to work with modifiers

### LINKED ISSUE #4336: bug: `list-col-wrap` doesn't work with `max-*` breakpoint ranges

https://play.tailwindcss.com/jOxzr23ZWL

_No response_

Safari

### Describe your issue

When `max-*` breakpoint ranges is used on the `list-col-wrap` the column doesn't wrap. As soon as breakpoint range is removed it starts to wrap (see tailwind play example).

pdanpdan:
example: https://play.tailwindcss.com/xI55YKx0Jz?file=css

close #4336


saadeghi:
Thank you @pdanpdan 🙏

PR #4338, opened 2025-12-11
fix(collapse): add transition for `min-height`

### LINKED ISSUE #2615: bug: Accordion not animating height

4.4.17

Chrome, Firefox

### Reproduction URL

https://daisyui.com/components/accordion/

### Describe your issue

Is it intentional that the accordion element does not transition it's height property? Opening the accordion element looks jerky, as only the padding transition is animated, and the content just pops in. I tested it out with setting transition: 0.2s ease (all props) and the result looks much smoother.

pdanpdan:
- firefox ignores grid-row height and transitions directly to full min-height
- delay start and shorten the duration of padding transition so that it is not so visible on the end of open/close

example: https://play.tailwindcss.com/9yLVxz7jaE?file=css

close #2615

### DIFF
--- a/packages/daisyui/src/components/collapse.css
+++ b/packages/daisyui/src/components/collapse.css
@@ -184,6 +184,7 @@
         content-visibility 0.2s allow-discrete,
         visibility 0.2s allow-discrete,
         min-height 0.2s ease-out allow-discrete,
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you 💚

PR #4342, opened 2025-12-12
fix(join): borders and outline for elements inside join

### LINKED ISSUE #4320: bug: Hover for buttons in groups is inconsistently highlighting borders

https://daisyui.com/components/join/

v5.5.5

Firefox

### Describe your issue

Example from daisyui.com. Right border/outline is not highlighted. Feels odd. Is it by design?
Image

pdanpdan:
- revert `z-index: 1` on input and select
- use the approach of increasing z-index for all direct children of `.join` that have focus inside. That way `.indicator` also works as expected (https://play.tailwindcss.com/LIjTRtOUTm?file=css).

The other approach was to increase z-index for `.join-item` that has focus inside, but it will break when using `.indicator` (https://play.tailwindcss.com/G9kR2UpdSr?file=css).

close #4320

### DIFF
--- a/packages/daisyui/src/components/input.css
+++ b/packages/daisyui/src/components/input.css
@@ -46,7 +46,6 @@
       outline: 2px solid var(--input-color);
       outline-offset: 2px;
       isolation: isolate;
-      z-index: 1;
     }
 
     /* increase font size in iOS so the page won't zoom */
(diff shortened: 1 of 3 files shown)


saadeghi:
Thank you @pdanpdan 🙏

PR #4355, opened 2025-12-19
docs(theme picker): colori does no know how to process hex so do not change the mode of the parsedColor but set the mode directly on colorState

pdanpdan:


pdanpdan:
@saadeghi please do not forget about this
It's a fix for a problem I introduced when refactoring theme builder

PR #4357, opened 2025-12-23
feat: add btn-bubble hover animation effect

codersjj:
Add .btn-bubble modifier class that creates an animated background bubble effect on button hover. The implementation includes:

- Smooth scale animation from center using CSS transforms
- Configurable animation properties via CSS custom properties:
  - --btn-bubble-duration (default: 350ms)
  - --btn-bubble-scale-x/y (default: 1.5/2.2)
  - --btn-bubble-easing (default: cubic-bezier)
  - --btn-bubble-bg for custom bubble color
- Automatic contrast preservation using color-mix()
- Accessibility support with prefers-reduced-motion
- Hardware acceleration with will-change: transform
- Compatible with all existing button variants and sizes
- Proper CSS layer integration (daisyui.l1.l2.l3)

The bubble effect inherits button border radius and maintains proper z-index layering behind button content. Includes comprehensive test suite with 39 passing tests covering property-based testing, accessibility, and cross-browser compatibility.

### DIFF
--- a/packages/daisyui/functions/buttonBubble.test.js
+++ b/packages/daisyui/functions/buttonBubble.test.js
@@ -1,0 +1,2626 @@
+import { expect, test } from "bun:test"
+import { JSDOM } from "jsdom"
+
+// Setup DOM environment for CSS testing
+const dom = new JSDOM(`<!DOCTYPE html><html><head></head><body></body></html>`)
+global.window = dom.window
+global.document = dom.window.document
+
+// Helper function to create a button element with bubble class
+function createBubbleButton(additionalClasses = "") {
+  const button = document.createElement("button")
+  button.className = `btn btn-bubble ${additionalClasses}`.trim()
+  button.textContent = "Test Button"
+  document.body.appendChild(button)
+  return button
+}
+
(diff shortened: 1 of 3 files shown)


saadeghi:
Thanks for the PR.  
This is an interesting effect but I'm not sure if it adds any value for the majority of daisyUI users.  

If anyone wants to add it their UI, it would be easy to add, but adding it to the library for everyone, wouldn't be in the scope of the design system.  

For future PRs, please open a GitHub Discussion first so we can decide together and confirm if a feature will be added or not, to avoid unwanted effort.  
Thanks 💚

codersjj:
> Thanks for the PR. This is an interesting effect but I'm not sure if it adds any value for the majority of daisyUI users.
> 
> If anyone wants to add it their UI, it would be easy to add, but adding it to the library for everyone, wouldn't be in the scope of the design system.
> 
> For future PRs, please open a GitHub Discussion first so we can decide together and confirm if a feature will be added or not, to avoid unwanted effort. Thanks 💚

Understood, thanks for the feedback! I'll make sure to start a GitHub Discussion for any future ideas first.

PR #4363, opened 2025-12-31
feat(tooltip): add alignment utility classes (start, center, end)

pdanpdan:
- apply to all positions (top, bottom, left, right)
- add a new variable --tt-tail-off to set offset of the tail for start and end alignment

example:  https://play.tailwindcss.com/Qt8vZTMlSe?file=css

close #4229

### DIFF
--- a/packages/daisyui/src/components/tooltip.css
+++ b/packages/daisyui/src/components/tooltip.css
@@ -105,7 +105,8 @@
   @layer daisyui.l1.l2 {
     > .tooltip-content,
     &[data-tip]:before {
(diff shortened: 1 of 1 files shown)


pdanpdan:
Maybe `tooltip-align-start`?

saadeghi:
I think `tooltip-tail-start` would be more understandable than `tooltip-align-start`.

It may not be clear what is `align` especially when the other class names are already changing the alignment from top, bottom, etc.

But `tail` is more clear. If we want the tail to be at the start, we would use `tooltip-tail-start`.

pdanpdan:
Are you changing the names or should I go with the rename?

saadeghi:
I will rename after merging to `v5.6` branch 👍  
I'm also adding `start` and `end` as aliases for `left` and `right`

PR #4366, opened 2026-01-05
Fix password input attribute from 'minlength' to 'minLength'

mirazahmeed:
What was changed
- Fixed invalid JSX attribute by replacing `minlength` with `minLength`.

Why
- JSX requires camelCase DOM attributes.
- Using `minlength` in JSX is ignored by React and breaks validation.


pdanpdan:
Hello. Please do not change it like that in the original html example. Take a look at https://github.com/saadeghi/daisyui/pull/4359 and add another replace case.

saadeghi:
Fixed in https://github.com/saadeghi/daisyui/pull/4369

PR #4368, opened 2026-01-05
feat(rating): make rating size modifiers responsive

### LINKED ISSUE #4367: bug: rating size classes do not have responsive classes

https://play.tailwindcss.com/LRoHG1UjGi

daisyui@5.5.14

All browsers

### Describe your issue

It looks like the responsive classes are not being generated for ratings sizes like "rating-md".
If you check the reproduction link you will see that md:rating-md does not work.

pdanpdan:
- simplify the CSS code to be based on size variables
- all sizes are based on --size-selector
- added rating-full (to override rating-half)

example: https://play.tailwindcss.com/GOQFYBDAnI?file=css

close #4367

### DIFF
--- a/packages/daisyui/src/components/rating.css
+++ b/packages/daisyui/src/components/rating.css
@@ -2,106 +2,89 @@
   @layer daisyui.l1.l2.l3 {
     @apply relative inline-flex align-middle;
 
-    & input {
-      border: none;
-      @apply appearance-none;
-    }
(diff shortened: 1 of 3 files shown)


saadeghi:
I think we don't need the new `rating-full` class. instead of `rating-half md:rating-full` we can use `max-md:rating-half`. Same result  
https://play.tailwindcss.com/y4Oxn8i8LS

I'll remove it in https://github.com/saadeghi/daisyui/pull/4579

PR #4371, opened 2026-01-06
feat(form elements): improve input, textarea, select

### LINKED ISSUE #4370: bug: wrong aspect fo input type datetime-locale on safari

https://play.tailwindcss.com/fVtLLFuSjO?file=css

latest

Safari

### Describe your issue

Picked from discord - added here to track the fix

pdanpdan:
- use size and font based on variables
- remove redundant rules
- fix input shadow element settings to work when using `.input input` and `input.input`
- fix placeholder color for safari
- make `.floating-label` respect responsive size modifiers on child `.input`, `.select`, `.textarea`

**Notable changes:**
- `.floating-label` is `flex` instead of `block` so that we can keep input `inline-flex` and avoid white space generated spacing
- input is `relative` so that the `::-webkit-calendar-picker-indicator` is correctly positioned when using `.input input`

**Not solved:**
- ~`.floating-label` does not adapt to responsive size classes of content (TODO in future PR)~

example (only for input): https://play.tailwindcss.com/s2xaEB1VZw?file=css

close #4370

### DIFF
--- a/packages/daisyui/src/components/input.css
+++ b/packages/daisyui/src/components/input.css
@@ -211,6 +211,11 @@
     --in-size-mul: 6;
     --font-size-min: 0.6875rem;
     --spin-my: -1;
+
+    .floating-label:has(&) {
+      --top-mul: 3;
+      --font-size: 0.6875rem;
+    }
   }
 }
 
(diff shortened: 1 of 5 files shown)


IdrisCelik:
Nice, would love to see this merged!

PR #4373, opened 2026-01-07
fix(prefix): replace only the variable name, not all content of `var(...)`

### LINKED ISSUE #4372: bug: rootcolor incorrectly generated when using DaisyUI prefix

https://play.tailwindcss.com/gvVolLYZsz

v5.5.14

Edge

### Describe your issue

I am using a prefix for DaisyUI. This is how my configuration looks:
```
@plugin "daisyui" {
  prefix: "d-";
}
```
Classes are correctly generated with the "d-" prefix, but there is an issue with the `rootcolor`.
The code for the `rootcolor` is generated as:
```
:root, [data-theme] {
  background: var(--d-page-scroll-bg, var(--root-bg)); // Here lies the problem.
  color: var(--color-base-content);
}
:where(:root, [data-theme]) {
  --d-root-bg: var(--color-base-100);
}
```

The fallback value for the root background is `--root-bg` instead of `--d-root-bg`.

### LINKED ISSUE #4408: bug: No prefix for `--page-has-scroll` variable

https://play.tailwindcss.com/Neh51twRYC

 v5.5.14

All browsers

### Describe your issue

Using prefix for DaisyUI classes content jumps right/left when open/close `modal`.

With `daisy-` prefix:
https://play.tailwindcss.com/Neh51twRYC
Without prefix:
https://play.tailwindcss.com/3O7hWvKpIR

When open `modal` using prefix like this:
```css
@plugin "daisyui" {
  prefix: "daisy-";
  themes: light --default;
}
```
there is result code with prefix but there is no prefix for `--page-has-scroll`, it should be `--daisy-page-has-scroll` in this case:
```css
:root {
  scrollbar-gutter: if(style(--page-has-scroll: 1): var(--daisy-page-scroll-gutter, unset);else: unset);
}
```

pdanpdan:
- existing regex was matching all text after `var(--` until the first `)`
- that solution does not work for `var` in `var` (`var(--a, var(--b))`)
- new implementation will only replace the variable name (`[a-zA-Z0-9_-]+`)
- it also accepts spaces between `var` and `(` and between `(` and `--`
- as this regex is heavily used I extracted it to be only compiled once

close #4372
close #4408

### DIFF
--- a/packages/daisyui/functions/addPrefix.js
+++ b/packages/daisyui/functions/addPrefix.js
@@ -154,13 +154,14 @@ const processArrayValue = (value, prefix, excludedPrefixes) => {
   })
 }
 
+const reVariableName = /var\s*\(\s*--([a-zA-Z0-9_-]+)/g
 const processStringValue = (value, prefix, excludedPrefixes) => {
   if (prefix === 0) return value
-  return value.replace(/var\(--([^)]+)\)/g, (match, variableName) => {
+  return value.replace(reVariableName, (match, variableName) => {
     if (shouldExcludeVariable(variableName, excludedPrefixes)) {
       return match
     }
-    return `var(--${prefix}${variableName})`
+    return `var(--${prefix}${variableName}`
   })
 }
 


kamontat:
@pdanpdan hi, may I ask what is blocking this to merge?

pdanpdan:
I think holiday/free time. Be patient :)

saadeghi:
Will be merged soon 👍

PR #4396, opened 2026-01-22
docs(dropdown): remove outdated compat info

samfundev:
Two fixes for the dropdown docs:
~1. The browser creates an implicit reference between the two elements when using `popovertarget`, so using `anchor-name`/`position-anchor` isn't necessary so the code can be simplified.~ Using the implicit reference causes [this bug](https://github.com/whatwg/html/issues/11694) so I've removed this change.
2. Firefox and Safari's latest release has support for CSS anchoring, so I've removed the section about compatibility.


saadeghi:
Thank you 🌼

PR #4415, opened 2026-02-02
fix: scope prose underline rule to a.btn only (fixes #4414)

### LINKED ISSUE #4414: Prose link underline shouldn't be removed
> Not sure if this was intended, but this also removes underline from the normal inline anchor text too: https://play.tailwindcss.com/4VrrcocUsK

_Originally posted by @ozdemirburak in https://github.com/saadeghi/daisyui/issues/4402#issuecomment-3837365143_

ozdemirburak:
Example: https://play.tailwindcss.com/NKJY7QNJZV?file=css

### DIFF
--- a/packages/daisyui/src/components/button.css
+++ b/packages/daisyui/src/components/button.css
@@ -4,7 +4,7 @@
   }
 }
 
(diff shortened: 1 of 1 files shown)


saadeghi:
Thank you @ozdemirburak

PR #4417, opened 2026-02-03
fix(card): only style figure if direct child of .card-body

### LINKED ISSUE #4416: bug: Diff component not working inside Card

https://play.tailwindcss.com/mWr6VLF5Fq

5.5.17

Firefox, Chrome, Other

### Describe your issue

**Description:**
The diff component does not render correctly when placed inside a card component. The issue appears to be caused by a conflict between the layer utilities, where the card layer overrides the properties of the diff layer.

**Expected behavior:**
The diff component should render correctly inside a card, just as it does outside of it.

**Actual behavior:**
The diff component is broken when inside a card (the slider and comparison functionality doesn't work).

**Possible solution:**
This might be resolved by adjusting the layer utility priorities or ensuring the diff component's styles take precedence when needed.

pdanpdan:
Initial approach, to be refined after further discussion (only style `figure` direct child of `.card`):

It looks like in card we apply styles to all `figure` inside, but all examples are with the images as direct child of `card` (and it woul make sense).

So I would say we should scope the style applied in card only to direct children of `card`.

https://play.tailwindcss.com/nZVaZs1bUq?file=css (has the example with the problem in #4416 and also all cards with image from docs)

What I am unsure about is the below, where my logic would say that `p` should also be only direct child of `card-body` (`> :where(p)`), but I am afraid that there are people that rely on the current behavior.

```css
card-body {
  @layer daisyui.l1.l2.l3 {
    @apply flex flex-auto flex-col gap-2;
    padding: var(--card-p, 1.5rem);
    font-size: var(--card-fs, 0.875rem);

    :where(p) {
      @apply grow;
    }
  }
}
```

close #4416

### DIFF
--- a/packages/daisyui/src/components/card.css
+++ b/packages/daisyui/src/components/card.css
@@ -13,7 +13,7 @@
       outline-color: currentColor;
     }
 
-    :where(figure:first-child) {
+    > :where(figure:first-child) {
       @apply overflow-hidden;
       border-start-start-radius: inherit;
       border-start-end-radius: inherit;
@@ -21,7 +21,7 @@
       border-end-end-radius: unset;
     }
 
-    :where(figure:last-child) {
+    > :where(figure:last-child) {
       @apply overflow-hidden;
       border-start-start-radius: unset;
       border-start-end-radius: unset;
@@ -29,7 +29,7 @@
       border-end-end-radius: inherit;
     }
 
-    figure {
+    > figure {
       @apply flex items-center justify-center;
     }
     &:has(> input:is(input[type="checkbox"], input[type="radio"])) {
@@ -66,7 +66,7 @@
       @apply text-neutral-content relative;
     }
 
-    :where(figure) {
+    > :where(figure) {
       @apply overflow-hidden;
       border-radius: inherit;
     }
@@ -174,7 +174,7 @@
     align-items: stretch;
     flex-direction: row;
 
-    :where(figure:first-child) {
+    > :where(figure:first-child) {
       @apply overflow-hidden;
       border-start-start-radius: inherit;
       border-start-end-radius: unset;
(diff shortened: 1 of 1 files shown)


pdanpdan:
Another approach would be to not style `figure` if it's a child of `.card-body`, but I would say it would make all very complex and error-prone.

saadeghi:
What if we change the layer of diff to `l2` and add `align-items: normal`?

https://play.tailwindcss.com/kiVnSotjoF?file=css

pdanpdan:
If the concern is only for `diff` in `card` then sure, it is a solution.
I was more concerned on the too broad selectors for `figure`.

I'm not sure if it is a good pattern to embed a card in a card, but something like this can happen: https://play.tailwindcss.com/nlNdv47z2d

saadeghi:
Right. Your solution is better for other situations as well 👍  
Will be merged this week

pdanpdan:
We should decide on the parent-child relationship for `p` - see first post

pdanpdan:
It looks like it does not fix as expected (#4474)
Maybe it's better to combine this with moving hover component to `l2`

saadeghi:
I don't think it's a good idea to make it strict fro the figure to be a direct child. It will be a breaking change if someone is putting the figure in an extra div (sometimes CMS nest extra divs and it's not even in control of the user)  

Instead of this PR, I changed the layers https://github.com/saadeghi/daisyui/commit/9a94fa81018f0b9407bbc9c33c4752151e08b244 that will fix https://github.com/saadeghi/daisyui/issues/4474 and https://github.com/saadeghi/daisyui/issues/4416  

It won't fix this situation https://play.tailwindcss.com/nlNdv47z2d but at least it's not a breaking change. Maybe we can fix that later using CSS scope.

Issue #4424, opened 2026-02-05
bug: list of problems observed on keyboard navigation


pdanpdan:
### Reproduction URL (Required)

https://daisyui.com/components/

### What version of daisyUI are you using?

latest

### Which browsers are you seeing the problem on?

All browsers

## https://daisyui.com/components/dropdown/#there-are-3-methods-to-use-a-dropdowns

- [ ] `h2` have no visual focus indicator (generic docs problem)
- [ ] `pre` with `tabindex` (code blocks) have no visual indicator

## https://daisyui.com/components/tab/#tabs-box

when the left tab has focus-visible the outline is below the next tab which is active
all other tab designs have inside outlines, so there is no problem

- [x] we should apply a `z-1`

## https://daisyui.com/components/filter/#filter-without-html-form

when no option is selected and the reset radio has focus there is no visual indicator

- [ ] we should show the reset option on focus-visible

## https://daisyui.com/components/drawer/#navbar-menu-for-desktop--sidebar-drawer-for-mobile

- [ ] example needs `drawer-button`
- [ ] the checkbox should be hidden when drawer is `drawer-open`

## https://daisyui.com/components/drawer/#responsive-collapsible-icon-only-drawer-sidebar-using-is-drawer-close-and-is-drawer-open

here the checkbox is hidden

- [ ] but in this case it should not be, because the button toggles the size of drawer
- [ ] these last 2 need a deeper analysis

PR #4425, opened 2026-02-05
fix(tab): apply `z-1` on box tab when focus-visible

pdanpdan:
when the left tab has focus-visible the outline is below the next tab which is active all other tab designs have inside outlines, so there is no problem

ref #4424


saadeghi:
Thank you @pdanpdan

PR #4427, opened 2026-02-06
feat(filter): show all options on keyboard navigation

pdanpdan:
- the last example in docs for filter does not use filter :) - it needs a fix for button because now button will not show as selected

example: https://play.tailwindcss.com/D4HeWvWziX

ref #4424


packages/docs/src/routes/(routes)/components/filter/+page.md
```diff
@@ -54,7 +54,7 @@ classnames:
 </div>
 ```
 ### ~Filter using HTML form, checkboxes, and a reset button
-#### For having multiple choices, use checkboxes. It doesn't need the filter class name.
+#### For having multiple choices, use checkboxes.
 
 <form>
```
pdanpdan, line 59: I think this should be `<form class="filter">` - it got lost in the merge with the changes from 5.6

saadeghi: Yes. Fixed it in a new commit.  
Thank you @pdanpdan 💚

PR #4428, opened 2026-02-06
feat(mockup-code): add optional .mockup-code-btn part class

codybrom:
Adds optional `.mockup-code-btn` part class for `.mockup-code`. Useful for copy buttons or other actions.

- Absolute positioned top-right with `z-10`
- Inherits parent text color with subtle `bg-current/10` background
- 60% opacity, transitions to 100% on hover
- CSS Only - Users have provide their own click handler.
- Docs updated with "With button" example and `part` classname entry

Example: https://play.tailwindcss.com/FIPJooQLyR?file=css

image

### DIFF
--- a/packages/daisyui/src/components/mockup.css
+++ b/packages/daisyui/src/components/mockup.css
@@ -13,6 +13,13 @@
         4.2em 0;
     }
 
+    .mockup-code-btn {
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks for the PR.  
Unfortunately this is not going to be merged. There are many customizations people can do to code blocks, such as positioning elements on top to show the language, to show a toggle, a copy button (like your example).   
And for a button, people may need different styles. I think it would be easier to let people choose the content they want to put there and not to force a specific one with a specific style.

PR #4432, opened 2026-02-08
feat(btn): full rewrite of button styles

### LINKED ISSUE #4430: bug: buttons with low contrast, aria-disabled support, buttons in filter

https://play.tailwindcss.com/xdk5JEP7oO

latest

All browsers

### Describe your issue

follow up from #4413

I have prepared 3 playgrounds that should cover most of the normal usage:
- 5.4.7: https://play.tailwindcss.com/ftj4zBp7HT
- 5.4.8+: https://play.tailwindcss.com/xdk5JEP7oO
- proposed: https://play.tailwindcss.com/GfY2pUJqSu

Problems:
- [ ] if buttons (checkbox) are used in `.filter` they do now show active status
addressed in #4427

- [ ] `aria-disabled="true"` is not accounted for
adressed in proposal by adding it to disabled states

- [ ] on all dark themes `.btn-neutral .btn-soft` cannot be used
addressed in proposal by introducing a variable to set the soft bg color for neutral (`--btn-soft-bg: var(--color-neutral-content) 80%;`)

- [ ] `.btn-soft`, `.btn-outline`, `.btn-dash`, `.btn-ghost`, `.btn-link` when there is contrast between fg (`--btn-fg`) and bg (`--btn-color`) button color, but there is no contrast in rest state between bg (derivative of `--color-base-100` or page bg) and fg (`--btn-color`)
addressed in proposal by introducing a user facing variable `--btn-rest-fg` that sets the fg color for buttons in rest state (only for soft, outline, dash, ghost) - it is ignored for `.btn-neutral` because that color has its own rules
`.btn-link` does not need any mitigation because it uses the same color all the time, so it can already be set by the user
in the example `--btn-rest-fg` is set for the  whole theme, but it can be used for individual buttons also

pdanpdan:
Organize styles so that there is very little need for duplication.
The only styles that do not work with tailwind responsive modifiers are the ones for disabled .btn-link and .btn-ghost.

I have prepared 3 playgrounds that should cover most of the normal usage:
- 5.4.7: https://play.tailwindcss.com/uUitizu33y?layout=preview
- 5.4.8+: https://play.tailwindcss.com/mkA67TyHx8?layout=preview
- full rewrite: https://play.tailwindcss.com/QZaRC8aifn?layout=preview

Problems fixed:
- [x] if buttons (checkbox) are used in `.filter` they do not show active status
  - also in #4427, so it would be better to first merge that one
- [x] `aria-disabled="true"` is not accounted for
- [x] on all dark themes `.btn-neutral .btn-soft` cannot be used
  - addressed by introducing an internal variable to set the soft bg color for neutral (`--int-soft-bg: var(--color-neutral-content) 80%;`)
- [x] `.btn-soft`, `.btn-outline`, `.btn-dash`, `.btn-ghost`, `.btn-link` when there is contrast between fg (`--btn-fg`) and bg (`--btn-color`) button colors, but there is no contrast in rest state between bg (derivative of `--color-base-100` or page bg) and fg (`--btn-color`)
  - new user facing variable `--btn-rest-fg` that sets the fg color for buttons in rest state (only for soft, outline, dash, ghost) - it is ignored for `.btn-neutral` because that color has its own rules
  - `.btn-link` does not need any mitigation because it uses the same color all the time, so it can already be set by the user
  - in the example `--btn-rest-fg` is set for the  whole theme, but it can be used for individual buttons also
- [x] a checkable button cannot be identified as checked/unchecked when focused
- [x] setting `--btn-color` and ``--btn-fg` on buttons does not work as expected

close #4430

### DIFF
--- a/packages/daisyui/src/components/button.css
+++ b/packages/daisyui/src/components/button.css
@@ -8,7 +8,6 @@
   @apply no-underline;
 }
 
-/* 1. BASE STRUCTURE */
 .btn {
   @layer daisyui.l1.l2.l3 {
     /* User-facing defaults */
@@ -19,7 +18,6 @@
     --btn-fg: var(--color-base-content);
     --btn-noise: var(--fx-noise);
 
-    /* Non-changing styles */
     @apply inline-flex shrink-0 cursor-pointer flex-nowrap items-center justify-center gap-1.5 text-center align-middle outline-offset-2 select-none;
     font-weight: 600;
     border-start-start-radius: var(--join-ss, var(--radius-field));
@@ -32,7 +30,6 @@
     transition-timing-function: cubic-bezier(0, 0, 0.2, 1);
     transition-duration: 0.2s;
 
-    /* Internal logic */
     --int-bg: var(--btn-color, var(--color-base-200));
     --int-fg: var(--btn-fg);
     --int-border: color-mix(
@@ -51,7 +48,6 @@
     --int-inset-def: 0 0.5px 0 0.5px oklch(100% 0 0 / calc(var(--depth) * 6%));
     --int-inset: var(--int-inset-def);
 
-    /* Apply internals */
     height: var(--size);
     padding-inline: var(--btn-p);
     font-size: var(--fontsize, 0.875rem);
@@ -62,7 +58,6 @@
     outline-color: var(--btn-color, var(--color-base-content));
     --tw-prose-links: var(--int-fg);
 
-    /* Effects */
     background-image: none, var(--int-noise);
     background-size: auto, calc(var(--noise, 0) * 100%);
     text-shadow: 0 0.5px oklch(100% 0 0 / calc(var(--depth) * 0.15));
@@ -70,7 +65,6 @@
       var(--int-inset) inset,
       var(--int-shadow);
 
-    /* Checkable label */
     &:is([type="checkbox"], [type="radio"]) {
       @apply appearance-none;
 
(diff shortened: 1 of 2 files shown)


saadeghi:
Thanks for the PR @pdanpdan 💚  

I think the new set of `--int-*` variables can get more minimal. All variables used in a specific component are already "internal" so I think adding more variables called `--int-*` would be confusing.  

Here https://github.com/saadeghi/daisyui/pull/4432/changes/dfb9fed0eccdf92d175ce5d0f69b764c3a81c5f6  
I renamed:
```
--int-bg -> --btn-bg
--int-border -> --btn-border
--int-soft-bg -> --btn-soft-bg
--int-shadow -> --btn-shadow
--int-inset -> --btn-inset
```
And removed:
```
--btn-noise
--int-fg
--int-border-style
--int-noise
--int-no-shadow
--int-shadow-def
--int-inset-def
```
Since most of them were not helping much.  
For example
```
--int-border-style: solid;
border-style: var(--int-border-style);
```
could be 
```
border-style: solid
```
The result seems to be same: https://play.tailwindcss.com/f2HbekQ9aq?file=css

```

size + number of variables:

│    │ file                           │ selector │ var   │ raw     │ brotli │ %  │
├────┼────────────────────────────────┼──────────┼───────┼─────────┼────────┼────┤
│ -- │ components/button.css          │ 228      │ 1104  │ 42.705  │ 1.889  │ 96 │ before this PR
│ -- │ components/button.css          │ 198      │ 1434  │ 43.665  │ 1.898  │ 96 │ before my last commit
│ -- │ components/button.css          │ 198      │ 1182  │ 43.425  │ 1.856  │ 96 │ after my last commit

pdanpdan:
`--int-border-style` was designed to allow having dashed border, but switch to solid border for hover/active states - it's no very easy to see, but on pressing a dashed button the border remains visible :)

Please let me check the real changes (what is not just renaming) to see if there is something else.


`--int-shadow-def` and `--int-inset-def` were used just to remove some long repetitive (and needing change in sync) values, so nothing functional there

I think the only problematic change (meaning hard to reason about) is the one about `--btn-fg` - now color is applied in multiple places and great care is needed for specificity/priority of rules.

But on visual inspection it looks good (except the border)

pdanpdan:
The last commit only fixed the dashed border on mouse hover (not on kbd) but i think it's ok

https://play.tailwindcss.com/niZgBO82ze?file=css

saadeghi:
Missed that. My bad.  
I will fix it in another commit 👍

PR #4441, opened 2026-02-20
Add OTP Input component

sulimanbenhalim:
New pure CSS OTP/PIN input component following daisyUI guidelines. No JavaScript required.

### DIFF
--- a/packages/daisyui/build.js
+++ b/packages/daisyui/build.js
@@ -102,6 +102,7 @@ async function generateFiles() {
           "filter",
           "mask",
           "mockup",
+          "otpinput",
           "skeleton",
           "swap",
           "validator",
--- a/packages/daisyui/src/components/otpinput.css
+++ b/packages/daisyui/src/components/otpinput.css
@@ -1,0 +1,208 @@
+.otp-input {
+  @layer daisyui.l1.l2.l3 {
+    --_w: 1ch;
+    --_g: 0px;
+    --_b: var(--border);
+    --input-color: color-mix(in oklab, var(--color-base-content) 20%, #0000);
+
+    box-sizing: content-box;
+    direction: ltr;
+    border: var(--_b) solid var(--input-color);
+    @apply bg-base-100 inline-block appearance-none;
+    font-family: var(--font-mono);
+    font-variant-numeric: tabular-nums;
+    font-size: max(var(--font-size, 1.25rem), 1.25rem);
+    line-height: 2.4;
+    letter-spacing: calc(var(--_w) + var(--_g));
+    width: calc(var(--digits) * (1ch + var(--_w) + var(--_g)));
+    padding-inline-start: calc(var(--_w) / 2);
+    padding-inline-end: 0;
+    clip-path: inset(0 calc((var(--_w) + var(--_g)) / 2) 0 0);
+    touch-action: manipulation;
+    border-start-start-radius: var(--join-ss, var(--radius-field));
+    border-start-end-radius: var(--join-se, var(--radius-field));
+    border-end-start-radius: var(--join-es, var(--radius-field));
+    border-end-end-radius: var(--join-ee, var(--radius-field));
+    box-shadow:
+      0 1px color-mix(in oklab, var(--input-color) calc(var(--depth) * 10%), #0000) inset,
+      0 -1px oklch(100% 0 0 / calc(var(--depth) * 0.1)) inset;
+
+    /* Internal vertical separators only — real CSS border handles the outer border */
+    background-image: repeating-linear-gradient(90deg,
+      var(--input-color) 0 var(--_b),
+      #0000 var(--_b) calc(1ch + var(--_w) + var(--_g))
+    );
+    background-size: calc(var(--digits) * (1ch + var(--_w) + var(--_g))) 100%;
+    background-position: calc(1ch + var(--_w) + var(--_g) / 2 - var(--_b) / 2) 0;
+    background-repeat: no-repeat;
+
+    &:focus,
+    &:focus-within {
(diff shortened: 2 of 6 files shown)


pdanpdan:
You can get away with less CSS if you design it as a modifier for `input`.

https://play.tailwindcss.com/zrB3E7tjAq?file=css

I left the outline on focus when it is in wrapper (label) but it can be removed and the same increased border size can be used for focus indication.

saadeghi:
I did some experiments on this a few months ago but I couldn't get the letter-spacing to work as expected. This looks promising 👍 Well done.

sulimanbenhalim:
> I did some experiments on this a few months ago but I couldn't get the letter-spacing to work as expected. This looks promising 👍 Well done.

still needs some work, it has issue on fully rounded themes like `valentine` worth looking into @pdanpdan solution

pdanpdan:
I managed to get rounded boxes for each char.
I would say that now the only problem is that we cannot use transparent colors for the border and the bg of the boxes (so focus/active state should be indicated using bg color or something else), but appart for this it obeys everything in the theme.

v1: ~https://play.tailwindcss.com/FMTa8yKr3I?file=css~
v2: ~https://play.tailwindcss.com/k0fWKxXvmB~ (added Valentine theme for testing, no other changes)
v3: https://play.tailwindcss.com/nxMhHDRqcE?file=css (improved aspect on text selection)

A variant with gap 0 should be separately defined, because it can be done with only one filled bg and repeated vertical lines, while using the input border and border radius for the external border.

sulimanbenhalim:
per-cell rounded boxes via 12-layer tiled gradient compositor

pdanpdan:
can you please put it in a play tailwind so we can check it?

sulimanbenhalim:
> Contributor

https://play.tailwindcss.com/K9X9n1TByk

pdanpdan:
It looks like there's still a problem with the connected variant - it scrolls after you insert all digits

sulimanbenhalim:
> It looks like there's still a problem with the connected variant - it scrolls after you insert all digits

Correct
I focused more on the separated variant and still couldn't get the weird looking glow off

saadeghi:
There's a UX issue, I'm not sure if there's a way to overcome it.  
Imagine the user typing one of the numbers wrong. They would click that part to remove it or edit it. Ideally clicking on number must put the cursor **after** the number so pressing the delete key on keyboard, deletes the number. But since we're using one input and we're using letter spacing, clicking a number puts the cursor **before** that number which is not ideal.

Video from [Vuetify OTP](https://vuetifyjs.com/en/components/otp-input/#usage) comparing to this PR, trying to remove the number **2**:

https://github.com/user-attachments/assets/b14a1192-3190-467e-b9a4-a8ce02362885

pdanpdan:
I'm not sure this can be fixed without js

sulimanbenhalim:
I don't think the whole thing is ready with the current CSS limitations.
also not sure if CSSWG have anything inthe pipeline that could solve it either

pdanpdan:
I think I have a much better and flexible solution:
- it uses the size and color modifiers from input
- it uses markup for boxes, so each one can be customized as desired
- on focus it goes to the end of input

~v1: https://play.tailwindcss.com/ud8HJowrJV~
v2: https://play.tailwindcss.com/EQZ1XhqFBs (font mono, small fixes for logical properties)

saadeghi:
@pdanpdan  I like this one more 👍  
It's less confusing for the end user,
also it allows the developer to customize the digit placeholders separately if needed.

I'm going to play with the style a little bit, maybe it could be a little bit simpler.

pdanpdan:
Just watch out that: 
- input needs pointer-events none to prevent focusing inside text
- input needs with and clamp to prevent scroll on last digit

Maybe renamed to input-otp
And I think ghost is not needed

sulimanbenhalim:
Cleanest approach so far, nice one 👍
The per-box markup makes it way more flexible!

One bug though: hitting spacebar counts the space as a char and shifts the rest of the digits one box over
Not sure if any CSS-only way to block that? pattern only fires on submit and type="number" kills leading zeros I guess...

Screenshot 2026-05-20 at 10 23 16 PM

pdanpdan:
I didn't think of that :)
It needs a mono font

sulimanbenhalim:
Ah yeah true, with mono at least the shift is obvious so the user sees something's off and hits backspace. Good enough :)

Screenshot 2026-05-20 at 10 31 57 PM

saadeghi:
Here's my suggestion:  
https://play.tailwindcss.com/Mt2lTGGgct

I removed the `underlined` variant. We don't have this style in daisyUI for other inputs.  
And the `ghost` variant. It was a different style from the existing [ghost style of daisyUI input](https://daisyui.com/components/input/#ghost-style), and that style won't be a good fit for OTP anyway, because of low contrast. It works for a normal input, because we may need an invisible input sometimes (for example in a navbar, with a search icon and a placeholder) but that style is not ideal for OTP. contrast is low and it won't work on different `base-*` colors.

I tried to simplify the style, I adjusted sizes so it fits the current scale of daisyUI (height of input = height of OTP).
I added variants for colors and sizes.  
Size variants are important for a responsive design,  
And colors are important because developers may need to to show error or success colors, or customize them the way they can customize other components.

pdanpdan:
Looks very nice.
I did some changes (commented with ` start `): https://play.tailwindcss.com/GkiAFSZYLb?file=css

For the outline color I'm not sure which version is better, so the changes are just so we can compare.

saadeghi:
Nice. Thank you 👍  
I opened a new PR https://github.com/saadeghi/daisyui/pull/4533

PR #4444, opened 2026-02-20
fix(breadcrumbs): use logical margins so that it works as expected in RTL

### LINKED ISSUE #4443: bug: Breadcrumb arrows wrong spacing in RTL

https://play.tailwindcss.com/PnsTWOKJxx?file=css

5.5.18

All browsers

### Describe your issue

Current Breadcrumbs RTL spacing:

Image

What should be:

Image

This issue was addressed in https://github.com/saadeghi/daisyui/issues/1598 before. But it seem that it never got fixed. The solution described in that issue, uses `[dir='rtl']` to fix the wrong `margin-left` and `margin-right` issues in RTL mode.

Now there are [`margin-inline-end` and `maring-inline-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/margin-inline-end) available with Baseline support.

Maybe we can use these new CSS margins instead that fixes the problem in RTL without separate selectors?

Also there is already a `[dir='rtl']` selector to rotate the breadcrumb arrows in RTL, we could also keep using `margin-left` and `margin-right` in there.

I'm happy to create a PR if that sounds OK (this is really triggering my OCD 😄).

pdanpdan:
close #4443

example: https://play.tailwindcss.com/Q5NwJppxTd?file=css

### DIFF
--- a/packages/daisyui/src/components/breadcrumbs.css
+++ b/packages/daisyui/src/components/breadcrumbs.css
@@ -25,7 +25,7 @@
 
         & + *:before {
           content: "";
-          @apply mr-3 ml-2 block h-1.5 w-1.5 opacity-40;
+          @apply me-3 ms-2 block h-1.5 w-1.5 opacity-40;
           rotate: 45deg;
           border-top: 1px solid;
           border-right: 1px solid;


saadeghi:
Thank you 🌼

PR #4445, opened 2026-02-20
feat: modernize rtl

pdanpdan:
- use `:dir(rtl)` instead of `[dir="rtl"]`
- use logical directions in a few places


saadeghi:
This can be a potential breaking change for some people, if they are nesting different directions. I need to do more testing before merging this

pdanpdan:
Yes, I agree - if they nested different directions it is a breaking fix :)
Maybe in the next minor release, because I would qualify this as a fix in order to get the expected behavior (and not as a changed behavior).

PR #4460, opened 2026-03-10
docs: add fixed height example to select

### LINKED ISSUE #4440: docs: Add example of select/dropdown with fixed height and many options

_No response_

### Describe the issue

It would be nice to have a docs example showing how to handle a large number of options with a fixed-height, scrollable menu, either using the `select` component or the `dropdown/menu` pattern.

All browsers

clint-code:
Added a fixed height usage example to the Select component documentation, by using the max-h- and overflow-y-auto utility classes, within the select menu, improving usability and layout stability.

Closes #4440

### DIFF
--- a/packages/docs/src/routes/(routes)/components/select/+page.md
+++ b/packages/docs/src/routes/(routes)/components/select/+page.md
@@ -104,7 +104,7 @@ classnames:
 
 <fieldset class="fieldset w-xs">
   <legend class="fieldset-legend">Browsers</legend>
-  <select class="select">
+  <select class="select max-h-60 overflow-y-auto z-[1]">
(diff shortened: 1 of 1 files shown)


saadeghi (changes requested):
Thanks for the PR.  

This example is "With fieldset and labels".  
Please add a new block for the fixed height, and make sure the `html` code block is also including the new class names.

pdanpdan:
Hello. Have you checked this? Can you please make a pen reproduction to show the result? Also the square brackets for z-1 are useless.

https://play.tailwindcss.com/BCdBXZllpk - it has no effect
check in the issue on how it should be done

PR #4461, opened 2026-03-10
feat: Add CSS vars --default-theme and --prefers-dark-theme

FabianoLothor:
Resolves #4457

### DIFF
--- a/packages/daisyui/functions/pluginOptionsHandler.js
+++ b/packages/daisyui/functions/pluginOptionsHandler.js
@@ -28,13 +28,18 @@ export const pluginOptionsHandler = (() => {
         if (flags.includes("--default")) {
           selector = `:where(${root}),${selector}`
         }
-        addBase({ [selector]: theme })
+        const themeWithFlags = { ...theme }
+        if (flags.includes("--default")) {
+          themeWithFlags["--default-theme"] = themeName
+        }
+        addBase({ [selector]: themeWithFlags })
 
         if (flags.includes("--prefersdark")) {
           // Use :root:not([data-theme]) for dark mode specificity
           const darkSelector =
             root === ":root" ? ":root:not([data-theme])" : `${root}:not([data-theme])`
-          addBase({ "@media (prefers-color-scheme: dark)": { [darkSelector]: theme } })
+          const darkThemeWithFlags = { ...theme, "--prefers-dark-theme": themeName }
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks for the PR.

You can add this variable in your own CSS.  
Why adding it to the library for everyone? 😅  

daisyUI is being used in half a million projects and countless users are using these projects. I don't think it's a good idea to add this for everyone.

FabianoLothor:
>  Why adding it to the library for everyone? 😅

Why other variables were added to everyone? Fuck this argument dude, bring something better to the table.

> You can add this variable in your own CSS.

Actually I can't declare this variable without declare the same theme name in two different places of the file, the idea of the PR is have a single source of truth.

```css
/* app.css */

@import 'tailwindcss';

@source not "./../vendor/daisyui/daisyui.index.js";

@plugin '@tailwindcss/typography';
@plugin './../vendor/daisyui/daisyui.index.js' {
    themes:
        wireframe --default, black --prefersdark
    ;
}
```

I am not in a node-bs-environment.

saadeghi:
> Can you help me? How can I achieve it in my CSS?

I don't know what's the goal, but here's a simple variable which you can later use with JS to get the theme names in the browsers:  
https://play.tailwindcss.com/0u2V58wymB?file=css

> Why other variables were added to everyone? 

Every single variable in the library is being used multiple times, deeply used in different themes and components, We use CSS variables for the values that need to be changed based on theme, based on component variants or based on customizations and all the existing variables are required. No variable is added without purpose.

> Fuck this argument dude, bring something better to the table.

I still don't know what is the objective. Maybe if you explain what (not how) you are trying to do, I can help.

> node-bs-environment 

What is that?

pdanpdan:
In the place where you define the options for daisyui you already know the names of the themes you are using. You can put them in a variable like you did in the pr.
But honestly it's easier to watch a match on prefers dark in js because you get the same information

FabianoLothor:
> > Can you help me? How can I achieve it in my CSS?
> 
> I don't know what's the goal, but here's a simple variable which you can later use with JS to get the theme names in the browsers: https://play.tailwindcss.com/0u2V58wymB?file=css
> 

So, this is fine, it is not the end of the world, a slightly different variation of that is my current workaround, but I was looking for something that would achieve this with a single source of truth.

When I change the value in the @plugin{} the CSS variable would already be aware of it.

Anyway, not the end of the world, I can find a different solution since the suggestion does not match with the goal of the library.

Sorry for the comment above, I flipped out.

PR #4480, opened 2026-03-27
docs: add examples for collapse usage

pdanpdan:
- collapse with checkbox and close on click outside
- collapse as expandable navbar container

preview: https://play.tailwindcss.com/UbyZvHbeWv
idea from #4479


saadeghi:
Thank you @pdanpdan 💚

PR #4499, opened 2026-04-09
docs: fix mockup-browser background color HTML snippet

pexisgle:
## Description
In the “browser mockup with background color” section, there was a discrepancy between the preview and the HTML snippet.  

## Changes
- Added the class `bg-base-100` to ensure the example accurately reflects the “background color” intent
- Adjusted the order of the classes to maintain consistency between the preview and the code snippet


saadeghi:
Thank you

PR #4519, opened 2026-05-10
fix: add missing tooltip-neutral color variant

### LINKED ISSUE #4492: bug: tooltip-neutral is not generated

https://play.tailwindcss.com/pVNAJeQmhX

v5.5.19

All browsers

### Describe your issue

Hi, `tooltip-neutral` is not working.

It doesn't generate a class for whatever reason.
All the others (tooltip-accent, ...) are working.

Image

`tooltip-neutral` also does not exist on autocomplete in tailwindplay

`tooltip-info` (and the others do show up)

`tooltip-neutral` is clearly missing.
It pretty much is a no-op
Image

OfirHaf:
## Problem

`tooltip-neutral` was never defined in `tooltip.css`, making it a silent no-op. Every other semantic color variant exists (`tooltip-primary`, `tooltip-secondary`, `tooltip-accent`, `tooltip-info`, `tooltip-success`, `tooltip-warning`, `tooltip-error`) but `neutral` was skipped.

This means:
- `tooltip-neutral` has no effect on styling
- It does not appear in autocomplete (Tailwind Play, IDE plugins)
- Users expecting consistent API across all color variants are left with a broken class

## Fix

Added `.tooltip-neutral` following the exact same pattern as the other color variants — sets `--tt-bg` to `var(--color-neutral)` and applies `text-neutral-content` to the tooltip content.

The placement in the file follows alphabetical ordering relative to the other variants.

Closes #4492

### DIFF
--- a/packages/daisyui/src/components/tooltip.css
+++ b/packages/daisyui/src/components/tooltip.css
@@ -191,6 +191,17 @@
   }
 }
 
+.tooltip-neutral {
+  @layer daisyui.l1.l2 {
+    --tt-bg: var(--color-neutral);
+
+    > .tooltip-content,
+    &[data-tip]:before {
+      @apply text-neutral-content;
+    }
+  }
+}
+
 .tooltip-error {
   @layer daisyui.l1.l2 {
     --tt-bg: var(--color-error);


saadeghi:
Hey, Thanks for the PR.  
Tooltip uses `neutral` color by default. So there would be no visual difference between `tooltip` and `tooltip tooltip-neutral`

OfirHaf:
Good point — you're right that `tooltip` already defaults to neutral. The main use case I was thinking of is resetting an inherited color modifier back to the default (e.g. inside a container that applies `tooltip-primary` to all tooltips), but I agree that's a pretty narrow edge case for a component library to handle explicitly. Happy to close this if you'd rather keep the API surface minimal.

saadeghi:
> inside a container that applies `tooltip-primary`

I wouldn't suggest doing that. And I won't guarantee that it will work in future versions 😅  
Use modifiers on the same element to modify them it will be easier to work with and easier to read the code.  

Thr goal for daisyUI is not to provide every class name for every color. We're trying to be more practical and be close to real life use cases.

Please let me know if you have any questions 🌼

OfirHaf:
Makes total sense, thanks for the clear explanation! Keeping modifiers on the same element is definitely cleaner. Appreciate you taking the time to review it 🌼

PR #4533, opened 2026-05-22
feat: add new OTP component

saadeghi:
Follow up to https://github.com/saadeghi/daisyui/pull/4441

### DIFF
--- a/packages/docs/src/routes/(routes)/components/otp/+page.md
+++ b/packages/docs/src/routes/(routes)/components/otp/+page.md
@@ -1,0 +1,301 @@
+---
+title: OTP
+desc: OTP (One-Time Password) component for inputting verification codes. It's usually 4 to 6 digits long and is used for two-factor authentication (2FA) or passwordless login.
+source: https://raw.githubusercontent.com/saadeghi/daisyui/refs/heads/master/packages/daisyui/src/components/otp.css
+layout: components
+classnames:
+  component:
+    - class: otp
+      desc: For the container label
+  size:
+  - class: otp-xs
+    desc: Extra small size
+  - class: otp-sm
+    desc: Small size
+  - class: otp-md
+    desc: Medium size
+    default: true
+  - class: otp-lg
+    desc: Large size
+  - class: otp-xl
+    desc: Extra large size
+---
+
+<script>
+  import Component from "$components/Component.svelte"
+  import Translate from "$components/Translate.svelte"
+</script>
+
+### ~OTP
+#### Make sure the number of spans matches the maxlength and the pattern of the input field. In this example, we have 4 spans for a 4-digit OTP.
+<label class="otp">
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <input type="text" autocomplete="one-time-code" inputmode="numeric" maxlength="4" pattern="[0-9]{4}" required />
+</label>
+
+```html
+<label class="$$otp">
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <input type="text" autocomplete="one-time-code" inputmode="numeric" maxlength="4" pattern="[0-9]{4}" required />
+</label>
+```
+
+### ~OTP with 6 digits
+
+<label class="otp">
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <input type="text" autocomplete="one-time-code" inputmode="numeric" maxlength="6" pattern="[0-9]{6}" required />
+</label>
+
+```html
+<label class="$$otp">
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <span></span>
+  <input type="text" autocomplete="one-time-code" inputmode="numeric" maxlength="6" pattern="[0-9]{6}" required />
+</label>
+```
+
+### ~OTP with connected boxes
+<label class="otp [--otp-gap:0rem]">
+  <span class="rounded-e-none border-e-0 outline-none"></span>
+  <span class="rounded-none border-e-0 outline-none"></span>
+  <span class="rounded-none border-e-0 outline-none"></span>
+  <span class="rounded-none border-e-0 outline-none"></span>
+  <span class="rounded-s-none outline-none"></span>
+  <input type="text" autocomplete="one-time-code" inputmode="numeric" maxlength="5" pattern="[0-9]{5}" required />
+</label>
+
+```html
+<label class="$$otp [--otp-gap:0rem]">
+  <span class="rounded-e-none border-e-0 outline-none"></span>
+  <span class="rounded-none border-e-0 outline-none"></span>
+  <span class="rounded-none border-e-0 outline-none"></span>
+  <span class="rounded-none border-e-0 outline-none"></span>
+  <span class="rounded-s-none outline-none"></span>
+  <input type="text" autocomplete="one-time-code" inputmode="numeric" maxlength="5" pattern="[0-9]{5}" required />
+</label>
+```
+
+### ~OTP with different sizes
+
+<div class="flex flex-col gap-4 items-center">
+  <label class="otp otp-xs">
+    <span></span>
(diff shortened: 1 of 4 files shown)


packages/docs/src/routes/(routes)/components/otp/+page.md
```diff
@@ -0,0 +1,301 @@
...
+    desc: Small size
+  - class: otp-md
+    desc: Medium size
+    default: true
+  - class: otp-lg
+    desc: Large size
+  - class: otp-xl
+    desc: Extra large size
+---
+
```
pdanpdan, line 43: I think color modifiers are missing

pdanpdan:
Suggestion to allow marking active box for cases where js is used (search for 'managed')
https://play.tailwindcss.com/a52Fj0eNDr

saadeghi:
@pdanpdan The feature is good visually 👍  
But I don't think we should use `aria-current="step"` here. For a screenreader, it's just a normal input and an empty span with `aria-current="step"` means nothing.  
Maybe a class name like `otp-current` instead. Or something else if you have a suggestion.

pdanpdan:
class name is ok with me

saadeghi:
Highlight the current character box with CSS only 😁  
https://play.tailwindcss.com/vRofvj9KRJ

- Using CSS `field-sizing` (Firefox doesn't support it yet, but it's ok, since it won't break the functionality)  
- Positions are a little bit buggy on Safari. I'm working on it…
- Works well on Chrome

pdanpdan:
Very interesting.
On chrome the only problem seems to be related to placeholder, which sets the width of the input to fit.

saadeghi:
I haven't seen placeholder being used for OTP.

pdanpdan:
Some tweaks: https://play.tailwindcss.com/8a73Y3ldsi
- does not require definition for all lengths
- looks good on ios
- can work with placeholder
- can work managed (I kept the attr, but can be chaged), with outline for all digits, with outline for next digit (if input is in div then auto)
- on firefox the div > input version does not show outline

saadeghi:
OTP doesn't need a placeholder.  
Some inputs like name, email, etc need placeholder, where the developer adds a `mail@site.com` as placeholder to show how the data must look like.  
You wouldn't want to write `1234` or `OTP` or `CODE` or… inside the OTP input.

And for the `managed` variant I prefer the CSS-only way. Since daisyUI does not rely on JS, and there's no universal piece of JS for this that would work on all frameworks.  

I'm working on the Safari issue for https://play.tailwindcss.com/vRofvj9KRJ and Firefox will support `field-sizing` soon (it's on Firefox Preview already)

pdanpdan:
I think the implementation with a wrapper div is cleaner from the pov of CSS.
It requires a wrapper div and an empty placeholder (this one to be able to detect empty OTP), but does not require fixed set of number of digits and allows using it in multiple ways: all boxes marked on focus / next box marked on focus

saadeghi:
> I think the implementation with a wrapper div is cleaner from the pov of CSS.

Which one?

pdanpdan:
https://github.com/saadeghi/daisyui/pull/4533#issuecomment-4534678636

If you remove the managed part is even smaller (I'll make the changes soon in an example) 
The one in link covers all 3 modes.


Version with managed version removed and with firefox fallback to highlight all boxes:
~https://play.tailwindcss.com/WEWxDlLKBU?file=css~
~https://play.tailwindcss.com/j73du1sFih - fixed the example~
https://play.tailwindcss.com/GHxu371zqP - I missed setting wrapper div flex (to prevent white space taking space)

It needs non-empty placeholder (" ")

saadeghi:
Thanks @pdanpdan 

My concerns about https://play.tailwindcss.com/GHxu371zqP:

1. here https://play.tailwindcss.com/GHxu371zqP when the input is full, the `:after` is still visible, highlighting a 5th character that is not supposed to visible.
Screenshot 2026-06-01 at 3 37 35 AM

2. if the purpose of `placeholder=" "` is to fix the positioning of the `:before` on the first character, I think we should fix it with some CSS adjustment under the hood. I think it's not appropriate to expect the developers to use an input with `placeholder=" "`. It will be removed by linters or any developer or LLM that is going to assume "I don't need a placeholder so this attribute is safe to remove".

3. having multiple syntaxes for as "managed", "All highlighted", "next highlighted" is not ideal. I understand that you added these versions for demo purpose (thanks), but at the end we only need one syntax. No JS, and the HTML as minimal as possible.

4. If we can get away with removing that `<div>` it would be nice.  
Even those empty spans are not ideal, but I couldn't find a way to get rid of them and having the same result (I tried painting rectangles with svg or CSS gradient as background, but that would be impossible to customize).  
I would suggest keeping it as minimal as possible: label + spans + input.

This is the latest demo from the current PR:  
https://play.tailwindcss.com/ImNYFYpnBG

I think not needing the `<div>` or `placeholder=" "` is worth it.

pdanpdan:
Ok. My main concern was related to the number of magic values in the CSS.
https://play.tailwindcss.com/4SUTpvr6cr - hide the caret on valid input (on iOS the caret is visible outside the OPT when filled)

saadeghi:
> Ok. My main concern was related to the number of magic values in the CSS.

I agree. That's not ideal. There's inconsistencies between Chrome and Safari, and in Safari with different zoom levels! I'm trying to fix those spacing bugs. Ideally when Safari fixes these bugs in future versions, the magic values won't be needed anymore.

I'm working on it to see if I can reduce those values 👍



> https://play.tailwindcss.com/4SUTpvr6cr - hide the caret on valid input

Thanks, I added to PR 🙏

pdanpdan:
Updated demo: https://play.tailwindcss.com/rwkeJJYPP9 (original)

A proposal to add --border with in paddin-inline start calculation (2 changes marked `here`): https://play.tailwindcss.com/UB4tcWK32e?file=css

sulimanbenhalim:
There is an issue with arabic-indic numerals

Screenshot 2026-06-04 at 12 06 45 PM

Maybe this can be fixed by adding "Kawkab Mono" font

saadeghi:
Does this font show English numbers as Arabic characters or are you writing Arabic numbers?

Because English and Arabic numbers are different characters in Unicode.

Also, do Arabic websites use Arabic numbers for OTP? Do you receive the SMS with Arabic numbers?

sulimanbenhalim:
No these are real Arabic-Indic characters being typed (different Unicode codepoints), not the font converting English numerals
and yes most Arabic sites send OTPs in English numerals, so this isn't really about Arabic-number OTPs

The realistic case is someone who left their keyboard on Arabic by mistake while typing the code: the CSS component looks broken to the user at first sight

unlike the JS OTP component handles the same input fine:

Screenshot 2026-06-04 at 12 22 21 PM

saadeghi:
>The realistic case is someone who left their keyboard on Arabic by mistake while typing the code.

In that case the backend should also handle this mistake by converting the numbers before validation.

I will try to fix this but I can't promise. Ideally OTP is expected to be English numbers. 

If my password is `abc123` no website would accept `abc۱۲۳` it's not the same.

sulimanbenhalim:
> backend should also handle this mistake

but that's not really what I'm raising, my concern is purely the UI. so if someone's keyboard is on Arabic and the boxes render broken like this, it reads as a buggy, poorly-built site to the user, even if the backend would handle the input fine.

That subliminal "this site is broken" impression is enough that I won't ship the CSS version for this

saadeghi:
I agree 👍

saadeghi:
> A proposal to add --border with in paddin-inline start calculation (2 changes marked `here`): https://play.tailwindcss.com/UB4tcWK32e?file=css

@pdanpdan Thanks. Added to PR.



> There is an issue with arabic-indic numerals

@sulimanbenhalim I couldn't find a solution for this. Chrome and Firefox completely ignore letter-spacing for Arabic characters. Safari is doesn't.

https://play.tailwindcss.com/0oBgyu0ONe?file=css

Apparently this change happened 5 years ago intentionally https://bugzilla.mozilla.org/show_bug.cgi?id=1656040

PR #4540, opened 2026-05-31
feat: add new Aura component

saadeghi:
demo https://play.tailwindcss.com/zHBsZVzFh7


pdanpdan:
Very nice.
I just updated the play to use `@utility` so that the last example works as expected: https://play.tailwindcss.com/D9CpnDMaEF

PR #4544, opened 2026-06-04
Update +page.md

tguelcan:
refactor: The command attribute for the button has an accompanying commandfor attribute, which can be an ID of an element. This means a native HTML <dialog> can be controlled entirely from HTML without needing any JavaScript.


saadeghi:
Thanks for the PR.  
I would love to merge, but we should wait until R***t supports this:  
https://github.com/facebook/react/issues/32478

PR #4547, opened 2026-06-10
fix(input): tweaks for safari quirks for datetime inputs

pdanpdan:
ref #4370

example: https://play.tailwindcss.com/xEdHm4ThAw

I tested it on ios and looks good
Can you please test on macos?


saadeghi:
> Can you please test on macos?

Looks good on MacOS Safari  👍  
Thank you

Issue #4549, opened 2026-06-11
bug: <select> overflows on chrome 149, does not respond to css rules


timotejroiko:
### Reproduction URL (Required)

https://play.tailwindcss.com/up35t4NFCN

### What version of daisyUI are you using?

5.5.23

### Which browsers are you seeing the problem on?

Chrome

### Describe your issue

Chrome 149 changed a lot of things regarding how select menus behave, especially with regards to `appearence:base-select`

Apparently the correct way of using it is like this now:

```html
<select class="select">
  <button>
    <selectedcontent style="white-space: nowrap; overflow: hidden"></selectedcontent>
  </button>
  <option>long text here<option>
  <option>long text h234234ere<option>
  <option>long text her23r2342342 234 234ere fwer werwer 23234 erwerwe rwer  234 234 234 23 42e<option>
</select>
```

what currently happens on the default select menu as shown on the daisyUI website:
<img width="987" height="356" alt="Image" src="https://github.com/user-attachments/assets/da57b4aa-b141-4d4d-94d1-8e4534e39262" />

This issue does not affect firefox

edsunman:
I think this is still an issue on latest daisyUI (version 5.6.10):

<img width="601" height="72" alt="Image" src="https://github.com/user-attachments/assets/e7126154-5107-403b-95d1-c7654c0c6eab" />

Chrome version: 149.0.7827.201 (Official Build) (64-bit)

pdanpdan:
You must use this syntax:

```html
<select class="select">
  <button>
    <selectedcontent></selectedcontent>
  </button>
  <option>long text here<option>
  <option>long text h234234ere<option>
  <option>long text her23r2342342 234 234ere fwer werwer 23234 erwerwe rwer  234 234 234 23 42e<option>
</select>
```

saadeghi:
The original syntax is also expected to look good.  
The new syntax is optional.

pdanpdan:
For the original syntax no styling can be applied, sorry

timotejroiko:
Issue is being tracked here:
https://issues.chromium.org/issues/521434907

Apparently some fixes are planned for chrome 151

However, transitioning to the new syntax seems to be the way forward and should probably be adopted soon

pdanpdan:
From the discord thread: 
For the moment the problem can be avoided using `appearance-none`
Using the new syntax for selectedcontent fixes the problem: https://play.tailwindcss.com/nQQpGfDaVX

[...]

PR #4550, opened 2026-06-12
feat: add default styling for selectedcontent slot

pdanpdan:
- add open indicator (rotated arrow)
- unify @supports block

Needs documentation update for button slot (I'll work on it)

example: ~https://play.tailwindcss.com/1ASzEP2k5N~
~https://play.tailwindcss.com/FjyCEMsJdH~ - add customized example
https://play.tailwindcss.com/YyuC0t2aIj - make example fallback on non-supporting browsers
ref #4549


saadeghi:
This can be merged, but it's too soon to add it to docs.  
Let's add it to docs after it gets a better support with frameworks.

PR #4551, opened 2026-06-13
feat: add new Megamenu component

saadeghi:
Demo:  
- https://play.tailwindcss.com/0Wfb8nL5ho
- https://play.tailwindcss.com/2cJC9W3F12
- https://play.tailwindcss.com/KqLGIBCMv3


pdanpdan:
Very nice.
Just the 3rd example needs overflow hidden because spring animation forces scrollbar shown on desktop

pdanpdan:
Do you think it's worth the idea to split this component in the mega menu and the active element indicator?

I'm thinking it would be nice to use in other situations also.

Something like https://github.com/saadeghi/daisyui/discussions/4412, but there it was mostly userland

saadeghi:
> I'm thinking it would be nice to use in other situations also.  

Yes but not with the same class names.  
Segment control can be a different class name with its own "active element indicator" even if the code is similar.  

I think in general it's a good idea to use them isolated, with different class names so when someone is using one component, there won't be any unused/unwanted styles.  
Right now a component can be excluded and the other components are expected to work without dependency to that component. For example variant class names like `select-primary` and `input-primary` are doing [mostly] similar things, but they work independently. Not very DRY, but it's more customizable and easier to work for the user.

saadeghi:
Maybe in the next major version we can update the tab component to have the transition effect. 
Or maybe segmented-control can be a new component. Hard to decide 😅

packages/daisyui/src/components/menu.css
```diff
@@ -156,7 +156,7 @@
 
 .menu-horizontal {
   @layer daisyui.l1.l2 {
-    @apply inline-flex flex-row items-center;
+    @apply inline-flex flex-row items-center; /* remove items-center*/
```
pdanpdan, line 159: Why this?

packages/daisyui/src/components/megamenu.css
```diff
@@ -0,0 +1,281 @@
+.megamenu {
+  @layer daisyui.l1.l2.l3 {
+    @apply relative flex items-center overflow-x-hidden;
```
pdanpdan: My bad about suggesting overflow hidden
It creates problems with outline of menu options
No setting for overflow defaults to auto -> scrollbars, and also in the navbar example hides the focus outline (horizontal parts)
But `overflow-visible` fixes the scrollbars and the outline

packages/daisyui/src/components/megamenu.css
```diff
@@ -0,0 +1,281 @@
...
+    &:has([popovertarget]:nth-of-type(9):hover):not(:has([popover]:popover-open)),
+    &:has([popover]:nth-of-type(9):popover-open) {
+      --mm-anchor: --mm9;
+    }
+    &:has([popovertarget]:nth-of-type(10):hover):not(:has([popover]:popover-open)),
+    &:has([popover]:nth-of-type(10):popover-open) {
+      --mm-anchor: --mm10;
+    }
+
+    &:has([popover]:nth-of-type(1):popover-open) [popovertarget]:nth-of-type(1):after,
```
pdanpdan: Can't we go with `[popovertarget]:has(+ [popover]:popover-open):after` to replace the whole list?
It would require the buttons and the panels to be interwoven (like in the documentation - so I would say it's ok)
Current solution would allow them to be all buttons grouped and all pannels grouped

packages/daisyui/src/components/megamenu.css
```diff
@@ -0,0 +1,281 @@
...
+    &:has([popover]:nth-of-type(2):popover-open) [popovertarget]:nth-of-type(2):after,
+    &:has([popover]:nth-of-type(3):popover-open) [popovertarget]:nth-of-type(3):after,
+    &:has([popover]:nth-of-type(4):popover-open) [popovertarget]:nth-of-type(4):after,
+    &:has([popover]:nth-of-type(5):popover-open) [popovertarget]:nth-of-type(5):after,
+    &:has([popover]:nth-of-type(6):popover-open) [popovertarget]:nth-of-type(6):after,
+    &:has([popover]:nth-of-type(7):popover-open) [popovertarget]:nth-of-type(7):after,
+    &:has([popover]:nth-of-type(8):popover-open) [popovertarget]:nth-of-type(8):after,
+    &:has([popover]:nth-of-type(9):popover-open) [popovertarget]:nth-of-type(9):after,
+    &:has([popover]:nth-of-type(10):popover-open) [popovertarget]:nth-of-type(10):after {
+      opacity: 1;
```
pdanpdan, line 120: I would also set `rotate: 45deg;` to ratate the arrow

packages/daisyui/src/components/megamenu.css
```diff
@@ -0,0 +1,281 @@
...
+    translate: 0 -0.5rem;
+    scale: 0.98;
+    transition:
+      opacity 200ms ease-out,
+      translate 200ms ease-out,
+      scale 200ms ease-out,
+      display 200ms ease-out allow-discrete,
+      overlay 200ms ease-out allow-discrete;
+    &:popover-open {
+      @apply fixed pt-4 opacity-100;
```
pdanpdan: `fixed` here produces side effects:
- if there is not enough space below it tries to align with the viewport so it's not opened just below the button but also it does not fit in page - see doc example with images)
- it does not stretch to fill content
- as a result the bg is transparent for the bottom part

without it I could not abserve any new side effect and the only side effect remaining is that if there is not enough space below it aligns to bottom of page (better than with viewport)

pdanpdan (commented):
I found some small problems - creating a PR with fixes

saadeghi:
Another issue:  

https://play.tailwindcss.com/z0H1XJlnIr

I wonder if there's a way to position an open popover as relative.


https://github.com/user-attachments/assets/50d716c2-e517-41c8-8da6-0d1cf2bec960

pdanpdan:
I noticed that problem too, but didn't dig into it yet. I'll check

There is no way to force the opened popover to exit top-layer.
What we could do is to force an opened popover megamenu to behave like a vertical one always

https://play.tailwindcss.com/19kRcEe0oc?file=css

This should be a refactored version with cleaned styles:
https://play.tailwindcss.com/joKXuSuggy?file=css

saadeghi:
I'm not sure, it's worse now 😅

pdanpdan:
Can you please add details about worse and now? :)
From my pov the last commit was interesting/nice - the only downside is that the menu height depends on the position of the activator on screen when opening

saadeghi:
On small screen, when I open the menu, then I change the screen size to big, it covers the whole page and there's no way to close it except pressing `Esc` key.  
It's because the anchor button that opened it is no longer visible (as expected)  
So user can't click the button, and can't click outside.

There must be a way to close it.  
Maybe, for this situation, we can let it cover ~40% of screen height, and add a `:backdrop` color (so it looks like a sheet). Then it would be obvious that clicking on the dark backdrop can close the popover.

Maybe something like
https://play.tailwindcss.com/iZdYMwijq9?file=css

---

I also tried another way: assigning an anchor name to each of navbar-start, navbar-center, navbar-end,
and position the open popover to that when this situation happens. but that method creates another challenge:  
if there are other elements before and after the megamenu, they will overlap. It will only work well, if there's only a mega menu in [navbar-start, navbar-center, navbar-end] and nothing else.

pdanpdan:
On android chrome fortunately you can use back button to close popovers :)
The play does not look bad - maybe limit it to 60-70% height (the main case is switching from portrait to landscape on mobile so you have very liitle vertical space) and vertical center?

saadeghi:
the fact that on this situation (open megamenu on small screen then resize to big), the whole menu became visible was weird.  
I made some changes, now, when this happens, only the first popover opens. This looks better now. it's the same size and position as how it would look like normally on a large screen. and it's more natural, because on a large screen, all popovers items are not supposed to be visible at the same time.  

I think it's good to merge now. @pdanpdan what do you think?  
https://play.tailwindcss.com/RXrVK6hQmm?file=css

pdanpdan:
The play looks good. Let me check the docs - brb

pdanpdan:
There is something wrong but I cannot identify the problem - in the docs for megamenu:
- open the menu on narrow screen
- increase the screen width
- the menu just dissapears

I cannot find the reason - i'll check more
The strange thing is that the same works in play

pdanpdan:
I found out why it looked like it worked in play :)
Because the menu is moved to the top of page, and in the play the navbar was on top

I added a margin top in the play so it is visible
https://play.tailwindcss.com/e9peTItcYT

It's not fixed - it's just adjusted to show the problem

saadeghi:
I'm not sure if I understood the issue.

The main menu (one, two, three buttons) will disappear. [same challenge](https://github.com/saadeghi/daisyui/pull/4551#issuecomment-4700129831). When the parent popover is open we can't position it relatively, so instead, to remind the user that the popover is still open and they should click outside to close it, I'm just showing the first child popover.

Also, the main menu (one, two, three buttons) is not relative as long as the main popover is open, so the first child popover cannot be positioned relative to that (the correct position, where the navbar is).  
I just positioned the first child popover on top of screen, where the navbar is expected to be.

So in [your example](https://play.tailwindcss.com/e9peTItcYT) if we remove the `mt-100`, it would be a good enough experience. Not perfect, but the best we can do.

I think it's good enough. Considering this situation is an edge case and it's not going to happen usually.

pdanpdan:
I was just describing the reason why in play it looks like it was working as expected, while in docs it was basically not working at all (because the popover is displayed at the top of the page, not of the screen)

pdanpdan:
I'll try 2 approaches:
- try to keep it as it is now and just fix the position (minimal CSS only changes): https://play.tailwindcss.com/Ms2Kg3l7bH
- ~wrap the activator button and the megamenu in a div so that it can be anchored in the right position when it is opened and not vertical - checking now if it is possible~ i can't find a good way for this, sorry

saadeghi:
Even if we can't find a way, it's ok. Megamenu is expected to be on the very top of the page.

saadeghi:
> * try to keep it as it is now and just fix the position (minimal CSS only changes): https://play.tailwindcss.com/Ms2Kg3l7bH

I'm going to keep this
```css
      > [popover]:not(:nth-of-type(1)) {
        display: var(--mm-display, block);
      }
```
as `none`:
```css
      > [popover]:not(:nth-of-type(1)) {
        display: var(--mm-display, none);
      }
```
so if the popover open on small screen and we change the screen size to large, it will only show the first child popover (not all of them)

other than that, I guess you're trying to make sure popover stays visible even if megamenu is in the middle (or bottom of page).  
If that's the case, it won't be an issue. Megamenu is expected to be at the top of page.   
But your changes are more bulletproof so I'm ok. I pushed the new changes 😁

saadeghi:
I'm merging this. If there are further improvements, we can push to `v5.6`.  Thanks @pdanpdan 💚

PR #4552, opened 2026-06-13
Fixes for megamenu - see review

pdanpdan:
- use overflow visible to prevent hiding outline and showing overscroll on megamenu-active spring mivement
- use --tw-content to allow customization
- simplify popover-open change of opacity selector
- for megamenu-vertical move apply always bg color so that on close it does not become transparent
- for megamenu-vertical  do not use fixed position
- replace some props with tw apply classes
- tailwind auto class ordering

- feat: rotate dropdown indicator when open


saadeghi:
Merged. latest demo: https://play.tailwindcss.com/AGcjX59uW4?file=css

PR #4554, opened 2026-06-14
feat: add vertical range slider (range-vertical)

### LINKED ISSUE #1483: feat: Vertical Range Slider

JamBalaya56562:
## Summary

Adds vertical range slider support, addressing #1483 (also attempted long ago in #3163, which predates the daisyUI 5 rewrite of the range component).

- `range-vertical` — renders the slider vertically (min at the bottom by default)

## Playground

▶️ https://play.tailwindcss.com/A7pEBBhKTa

Shows horizontal, vertical, plus sizes & colors.

## How it works

The daisyUI 5 range fill is painted by the thumb's last `box-shadow` layer (clipped by `overflow: hidden`). This change:

- Extracts that layer into `--range-fill-x` / `--range-fill-y` / `--range-fill-spread` so the fill can be repointed from the x axis to the y axis. The horizontal behavior is unchanged — the defaults reproduce the exact same computed values.
- Uses native `writing-mode` + `direction` for orientation instead of a `transform: rotate()` hack, so keyboard interaction, focus, and the accessibility tree all work natively.
- Swaps the Firefox `::-moz-range-track` width/height in vertical mode so the unfilled groove spans the full height (WebKit/Chromium already render it correctly via `::-webkit-slider-runnable-track`).

## Usage

```html
<!-- horizontal -->
<input type="range" min="0" max="100" value="40" class="range" />

<!-- vertical -->
<input type="range" min="0" max="100" value="40" aria-orientation="vertical" class="range range-vertical" />
```

## Testing

Rendered and verified across all three engines (Playwright + local preview):

- **Chromium**, **Firefox 150**, and **WebKit 26.4 (Safari engine)** — thumb position and fill direction/proportion are correct for horizontal (LTR & RTL), vertical, at values 0 / 25 / 50 / 75 / 100.
- Horizontal range: no regression from the fill-variable refactor.
- `aria-orientation="vertical"` added to the docs examples per the WAI-ARIA APG slider pattern.

Closes #1483

### DIFF
--- a/packages/daisyui/src/components/range.css
+++ b/packages/daisyui/src/components/range.css
@@ -8,6 +8,15 @@
     --range-fill: 1;
     --range-p: 0.25rem;
     --range-bg: color-mix(in oklab, currentColor 10%, #0000);
+    /* Horizontal fill direction: 1 for ltr, -1 for rtl (set in the [dir=rtl] block below). */
+    --range-dir: 1;
+    /* Fill is painted by the thumb's last box-shadow layer, clipped by overflow:hidden.
+       These variables let .range-vertical swap the fill from the x axis to the y axis. */
+    --range-fill-x: calc(
+      (var(--range-dir, 1) * -100cqw) - (var(--range-dir, 1) * var(--range-thumb-size) / 2)
+    );
+    --range-fill-y: 0;
+    --range-fill-spread: calc(100cqw * var(--range-fill));
     @apply cursor-pointer overflow-hidden bg-transparent align-middle;
     width: clamp(3rem, 20rem, 100%);
     /* --radius-selector-max is a separate variable because ~ calc(min(calc(--var))) ~ gives build error in PostCSS+Nuxt */
@@ -52,7 +61,7 @@
       border: var(--range-p) solid;
       appearance: none;
       -webkit-appearance: none;
-      top: 50%;
+      inset-block-start: 50%;
       color: var(--range-progress);
       transform: translateY(-50%);
       box-shadow:
@@ -60,8 +69,7 @@
         0 8px 0 -4px oklch(100% 0 0 / calc(var(--depth) * 0.1)) inset,
         0 1px color-mix(in oklab, currentColor calc(var(--depth) * 10%), #0000),
         0 0 0 2rem var(--range-thumb) inset,
-        calc((var(--range-dir, 1) * -100cqw) - (var(--range-dir, 1) * var(--range-thumb-size) / 2))
-          0 0 calc(100cqw * var(--range-fill));
+        var(--range-fill-x) var(--range-fill-y) 0 var(--range-fill-spread);
     }
 
     &::-moz-range-track {
@@ -83,15 +91,13 @@
       height: var(--range-thumb-size);
       width: var(--range-thumb-size);
       border: var(--range-p) solid;
-      top: 50%;
       color: var(--range-progress);
       box-shadow:
         0 -1px oklch(0% 0 0 / calc(var(--depth) * 0.1)) inset,
         0 8px 0 -4px oklch(100% 0 0 / calc(var(--depth) * 0.1)) inset,
         0 1px color-mix(in oklab, currentColor calc(var(--depth) * 10%), #0000),
         0 0 0 2rem var(--range-thumb) inset,
-        calc((var(--range-dir, 1) * -100cqw) - (var(--range-dir, 1) * var(--range-thumb-size) / 2))
-          0 0 calc(100cqw * var(--range-fill));
+        var(--range-fill-x) var(--range-fill-y) 0 var(--range-fill-spread);
     }
     &:disabled {
       @apply cursor-not-allowed opacity-30;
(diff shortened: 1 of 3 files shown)


pdanpdan:
Can you please add a playground with example to see how it works?

pdanpdan:
Also it would be nice: 
- default vertical to have min on bottom and max on top
- range-invert to also work for horizontal range (accounting for RTL)

JamBalaya56562:
Done — thanks for the feedback! 🙌

Playground with all variants: https://play.tailwindcss.com/g2wfksu9Vq

It covers:
- Horizontal `range-invert` (LTR) — fills from the opposite side
- Horizontal `range-invert` inside an RTL container — flips the other way
- `range-vertical` (default: min at the bottom) and `range-vertical range-invert` (min at the top)
- Sizes & colors

Re: your suggestions:
- The default vertical already goes **min at the bottom → max at the top** 👍
- I pushed a change so `range-invert` now also reverses **horizontal** ranges, RTL-aware — the native slider direction and the cosmetic fill flip together for every LTR/RTL + invert combination. Verified in Firefox and WebKit.

saadeghi:
vertical ones work correctly on Firefox, but on Chrome and Safari, they have -50% error.  
when value is 50% it shows 0  
when it's 100%, it shows 50%

saadeghi:
Also, does the invert variant have any usecase?  
Because I don't remember ever seeing an RTL range slider on a LTR page (or vice versa). Or an upside down vertical range slider 😅

JamBalaya56562:
Fixed the vertical fill offset on Chrome/Safari — thanks for catching it! 🙏

**Fix:** reset `top`/`transform` on the WebKit thumb inside `.range-vertical` only (horizontal ranges are untouched). I re-verified the fill is proportional at 0 / 25 / 50 / 75 / 100 in Chromium and Firefox.

**Re: `range-invert`** — fair question. I haven't seen a real use for an RTL-on-LTR (or upside-down vertical) slider either, so I'm inclined to drop `range-invert` and keep just `range-vertical`. Since @pdanpdan specifically requested the horizontal invert, I'd like to confirm with them before removing it — happy to take it out if there are no objections.

pdanpdan:
Example with some fixes:
- if we use `inset-block-start` instead of `top` for `::-webkit-slider-thumb` then we only need to change the translation axis for vertical
- the `top` value for `::-moz-range-thumb` is not needed (it does nothing anyway)
- added `::-webkit-slider-runnable-track` for vertical (it was not switching track dimensions for vertical)
- removed the `range-invert` and as such simplified some calculations
- modified to have minimal changes from original code

https://play.tailwindcss.com/2D9Ss2ukZA?file=css

pdanpdan:
About the modifier to invert range - the request was mainly for consistency (it was strange to only work on vertical), but I don't have real live usage examples :D

JamBalaya56562:
Thanks @pdanpdan, great cleanup — adopted your version. I verified each point in Chromium and Firefox (Playwright pixel tests):

- `inset-block-start` + swapping only the transform axis for vertical — works; horizontal LTR/RTL unchanged.
- The `top` on `::-moz-range-thumb` was indeed dead (removing it changes nothing in Firefox).
- Good catch on `::-webkit-slider-runnable-track` — the previous code rendered the unfilled vertical groove full-width on Chromium; switching the track dimensions restores the thin centered groove.
- I kept the `::-moz-range-thumb` block itself (Firefox needs it for the thumb sizing + box-shadow fill — removing the whole block drops the fill entirely); only the dead `top` line is gone.
- Dropped `range-invert` and its docs/skill entries, and collapsed the direction vars to a single `--range-dir`.

Pushed. 🙏

pdanpdan:
I think either `--range-dir: 1;` in `.range` should be removed or the fallback value in line 16 should be removed

@saadeghi do you have a preference?

saadeghi:
> I think either `--range-dir: 1;` in `.range` should be removed or the fallback value in line 16 should be removed

@pdanpdan Correct, `--range-dir: 1;` is not needed, because it has fallback.  

@JamBalaya56562 I wasn't able to add a commit to this PR, I opened a PR to your fork https://github.com/JamBalaya56562/daisyui/pull/1

Also, I removed the `aria-orientation` from docs. It's not for a native HTML input.  
It's for when we create custom range sliders using divs and spans and JS. And they would need a [`role="slider"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/slider_role) and [`aria-orientation`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation) attribute.

Native HTML range slider doesn't need any of them because it's already understandable and accessible to the browser, keyboard navigation, and screenreader.  
- https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-orientation
- https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/input/range#creating_vertical_range_controls

JamBalaya56562:
@saadeghi Thanks!
I have merged your PR.

saadeghi:
Merged to v5.6 branch ✅

PR #4555, opened 2026-06-15
feat: add vertical range slider (range-vertical & range-invert)

JamBalaya56562:
Add `range-vertical` for vertical sliders and `range-invert` to reverse the
slider direction. Built on native `writing-mode`/`direction`, so keyboard
interaction and accessibility work without a rotate hack.

- The fill is repointed to the vertical axis through new `--range-fill-x/y/spread`
  variables; the horizontal behavior is unchanged.
- `range-invert` is a standalone class (driven by custom properties) so
  responsive variants like `md:range-invert` are generated correctly. It
  reverses both vertical and horizontal sliders and is RTL-aware on horizontal:
  the native slider direction and the cosmetic fill flip together for every
  LTR/RTL + invert combination.
- In vertical mode, reset the WebKit thumb's `top`/`transform` (the horizontal
  cross-axis centering) so it no longer shifts the thumb along the travel axis.
  This fixes the ~50% fill offset on Chrome/Safari (value 50% rendered as 0,
  100% as 50%); Firefox ignores top/transform on the thumb and was unaffected.
- Firefox `::-moz-range-track` dimensions are swapped in vertical mode so the
  unfilled groove spans the full height.

Verified in Chromium, Firefox and WebKit.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/packages/daisyui/src/components/range.css
+++ b/packages/daisyui/src/components/range.css
@@ -8,6 +8,18 @@
     --range-fill: 1;
     --range-p: 0.25rem;
     --range-bg: color-mix(in oklab, currentColor 10%, #0000);
+    /* Horizontal fill direction. --range-rtl is the page direction (1 ltr, -1 rtl)
+       and --range-invert-sign is flipped by range-invert, so the fill mirrors
+       correctly for every LTR/RTL + invert combination. */
+    --range-rtl: 1;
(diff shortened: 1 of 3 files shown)


JamBalaya56562:
Accidental duplicate of #4554 (tooling created a new PR instead of updating the existing one). The fix has been pushed to #4554 directly. Closing this.

PR #4560, opened 2026-06-17
fix(fab): fan out the first fab-flower action when no main-action is set

### LINKED ISSUE #4531: bug: FAB Flower forces a main action button

https://play.tailwindcss.com/ZAXQ7WbusF

v5.5.20

Chrome

### Describe your issue

If you use the FAB in speed dial configuration, the actions appear above the main button.

Image

However, if you add the `fab-flower` class, the `A` button replaces the trigger button.

Image

I would expect this to only happen if the user specifically adds a `fab-main-action` class to it.

JamBalaya56562:
### Problem

In `.fab-flower` the center pin was `> *:nth-child(-n + 2) { --position: 0rem; }`, which pinned **both** the trigger (1st child) **and** the first action (2nd child) to the center. So when no `.fab-main-action` / `.fab-close` is present, the first action button overlaps and "replaces" the trigger instead of fanning out into the quarter-circle arc.

As reported by @topherfangio, this should only happen when a button is explicitly marked `.fab-main-action`. (@saadeghi: *"You're right. Will be fixed in v5.6 soon."*)

### Fix (minimal / additive)

Following review (@pdanpdan, @saadeghi) the change is kept as small and additive as possible:

1. **Center by identity, not position** — pin the trigger plus `.fab-main-action` / `.fab-close` (`> *:nth-child(1), > .fab-main-action, > .fab-close`).
2. **The existing arc blocks (`:has(:nth-child(3..6))`) are left untouched** — Mode A (with a main-action/close) is byte-identical to `v5.6`, so there is no regression risk.
3. A single block is **appended** for the no-main-action case: `&:not(:has(.fab-main-action, .fab-close))`, mirroring the existing degrees shifted by one child (the arc starts at the 2nd child) and hiding `:nth-child(n + 6)` so it caps at the same 4 actions as Mode A.

Net diff vs `v5.6` is just the identity-pin line + the appended `:not(:has())` block. (An earlier revision wrapped the existing rules in `:has(.fab-main-action)/:has(.fab-close)`; that wrapping has been reverted.)

### Playground

▶️ **Before / after, side by side:** https://play.tailwindcss.com/LifJWa8hZG

Rows are action counts, left column is the released (buggy) daisyUI, right column is this PR. Heads-up: Play loads the **released** daisyUI, so the right column re-creates the fix's result with a **clearly-labelled preview shim** on top of it — that shim is more verbose than this PR (it has to override the released `@layer` rules). The real CSS is the small diff above.

### Verification

`bun run build` succeeds (lightningcss accepts the `:not(:has())` selectors) and `bun test` passes (86/86).

Headless-Chromium measurement of each action's center offset from the trigger (button = 48px), opening each FAB via `:focus-within`:

| case | trigger | center button | arc actions (dist @ angle) |
|------|---------|---------------|----------------------------|
| no main / 1 | 0px | — | A 67px @135° |
| no main / 2 | 0px | — | A, B 67px @165° / 105° |
| no main / 3 | 0px | — | A, B, C 86px @180° / 135° / 90° |
| no main / 4 | 0px | — | A, B, C, D 106px @180° / 150° / 120° / 90° |
| `fab-main-action` / 3 | 0px | **0px (centered)** | A, B, C 86px @180° / 135° / 90° |
| `fab-close` / 3 | 0px | **0px (centered)** | A, B, C 86px @180° / 135° / 90° |

- **Without a main-action** the first action goes from overlapping the trigger (`dist 0`) to fanning out (`dist > 0`) at the correct angles.
- **With `fab-main-action` / `fab-close`** the center button stays on the trigger and the arc matches the unchanged `v5.6` geometry.

### Notes

- Single-component change in `packages/daisyui/src/components/fab.css`, plus a docs example (fab-flower without a main action) and a capacity note.

### DIFF
--- a/packages/daisyui/src/components/fab.css
+++ b/packages/daisyui/src/components/fab.css
@@ -76,7 +76,11 @@
   @layer daisyui.l1.l2.l3 {
     @apply grid;
     --position: 0rem;
-    > *:nth-child(-n + 2) {
+    /* Center the trigger and an explicit main-action/close by identity (not by
+       position), so the first real action fans out when neither is present. */
+    > *:nth-child(1),
+    > .fab-main-action,
+    > .fab-close {
       --position: 0rem;
     }
     > * {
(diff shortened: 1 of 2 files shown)


pdanpdan:
Hello.

I would say that the solution is a little bit convoluted.
I would go with:
- apply `--position: 0rem;` only to the FAB button and .fab-main-action and .fab-close (like in PR)
- keep the current CSS for flower
- at the end add the styles for `&:not(:has(.fab-main-action, .fab-close))` in the same way they are done above - only change the number of children
- also for this case `> :nth-child(n + 6)` should be hidden

saadeghi:
Thanks for the PR.  
I would appreciate if the CSS rule changes stay as minimum as possible. When this many lines are changing in a PR, we have to predict and test every possible situation to make sure it's not breaking existing apps in the next update.

JamBalaya56562:
Thanks @pdanpdan and @saadeghi — updated to keep the change minimal and additive, exactly along the lines you suggested:

- Kept the identity-based center pin (`> *:nth-child(1), > .fab-main-action, > .fab-close`).
- **Reverted the wrapping of the existing arc blocks** — they are now byte-identical to `v5.6`, so Mode A (with a `fab-main-action` / `fab-close`) is completely untouched and carries no regression risk.
- Appended a single `&:not(:has(.fab-main-action, .fab-close))` block, mirroring the existing degrees with the child numbers shifted by one, plus `> :nth-child(n + 6)` hidden (so the no-main case caps at the same 4 actions as Mode A).

So the net diff vs `v5.6` is now just the identity-pin line + that one appended block. Pushed as a new commit, and the description has the headless before/after numbers (no-main fans out, with-main/close stays centered and unchanged). 🙏

Issue #4575, opened 2026-06-23
bug: The modal's animation is interrupted when another modal is open


oscar370:
### Reproduction URL (Required)

https://play.tailwindcss.com/qcRITtPbyd

### What version of daisyUI are you using?

5.5.23

### Which browsers are you seeing the problem on?

Chrome

### Describe your issue

Calling `close()` instantly removes the modal from the `top-layer`. Without transitioning the overlay property, the closing modal falls back to the normal DOM flow immediately. This causes it to be instantly hidden behind the remaining open modal and its backdrop, which still occupy the `top-layer`.

In the attached Tailwind Play, I show what it looks like with the current version of Daisy and how to fix it by animating the overlay.

```css
/* Current */
transition: visibility 0.3s allow-discrete, background-color 0.3s ease-out, opacity 0.1s ease-out;

/* Proposed */
transition: overlay 0.3s allow-discrete, visibility 0.3s allow-discrete, background-color 0.3s ease-out, opacity 0.1s ease-out;
```

pdanpdan:
Nice catch. Can you please create a PR for this?

oscar370:
> Nice catch. Can you please create a PR for this?

Sure

PR #4576, opened 2026-06-23
fix(modal): add overlay transition to prevent abrupt exit on stacked dialogs

oscar370:
Adding a transition to the overlay property allows the dialog to be animated out when another modal is open.

This PR addresses the issue mentioned in: https://github.com/saadeghi/daisyui/issues/4575


saadeghi:
Thank you. merged to `v5.6` branch ✅

PR #4577, opened 2026-06-24
fix(btn): border style for keyboard and mouse navigation

pdanpdan:
example: https://play.tailwindcss.com/3GmlbzBseZ

follow-up from #4432


saadeghi:
Thank you 💚

PR #4582, opened 2026-06-25
docs: remove wrong modal-bottom class from modal-box in theme config

pdanpdan:
can be observed on narrow screens in CSS box


saadeghi:
Thank you

PR #4603, opened 2026-07-02
fix: #4602 indicator center and middle misalign when item is a child …

### LINKED ISSUE #4602: bug: indicator misaligned for center and middle placements when child in indicator-item

https://play.tailwindcss.com/O0HYtHwACc

v5.6.7

All browsers, Safari iOS, Firefox

### Describe your issue

On the demo page, the last example shows [a button as an indicator for a card](https://daisyui.com/components/indicator/#a-button-as-an-indicator-for-a-card). This appears to work until you add either `indicator-center` or `indicator-middle` to the element. The result is a misaligned element.

Please see the Tailwind Play link for more examples.

The bug comes from using `--indicator-{e/b}: 50%` instead of `--indicator-{e-b}: auto` in the `indicator-{center/middle}` elements respectively.

Providing this change does not break the appearance for those that do not nest elements in the html element tagged with `indicator-item`.

If I have misunderstood something / am misusing daisyUI to expose this bug please let me know.

eligrubbs:
…html tag.

closes: #4602

### DIFF
--- a/packages/daisyui/src/components/indicator.css
+++ b/packages/daisyui/src/components/indicator.css
@@ -30,7 +30,7 @@
 .indicator-center {
   @layer daisyui.l1.l2 {
     --indicator-s: 50%;
-    --indicator-e: 50%;
+    --indicator-e: auto;
     --indicator-x: -50%;
     [dir="rtl"] & {
       --indicator-x: 50%;
@@ -62,7 +62,7 @@
 .indicator-middle {
   @layer daisyui.l1.l2 {
     --indicator-t: 50%;
-    --indicator-b: 50%;
+    --indicator-b: auto;
     --indicator-y: -50%;
   }
 }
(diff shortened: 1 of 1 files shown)


packages/daisyui/src/components/indicator.css
```diff
@@ -30,7 +30,7 @@
 .indicator-center {
   @layer daisyui.l1.l2 {
     --indicator-s: 50%;
-    --indicator-e: 50%;
+    --indicator-e: auto;
     --indicator-x: -50%;
     [dir="rtl"] & {
```
saadeghi, line 35: Thanks for the PR 💚  
Can you please add this for the `[dir="rtl"]`?
```
--indicator-e: 50%;
```
Before: https://play.tailwindcss.com/T4rKzG9U4K?file=css  
After: https://play.tailwindcss.com/vghCAIVUmA?file=css

```diff
  .indicator-center {
    @layer daisyui.l1.l2 {
      --indicator-s: 50%;
      --indicator-e: auto;
      --indicator-x: -50%;
      [dir="rtl"] & {
+       --indicator-e: 50%;
        --indicator-x: 50%;
      }
    }
  }
```

saadeghi (changes requested):

PR #4608, opened 2026-07-03
fix(drawer): keep drawer-toggle keyboard-focusable when drawer is open

JamBalaya56562:
## What & why

Follow-up to #4607, addressing the remaining **responsive drawer** item from #4424.

The **responsive collapsible icon-only drawer** example stays open on large screens via `lg:drawer-open` and uses its button to toggle the sidebar **size** (icon-only ↔ expanded, via the `is-drawer-open:` / `is-drawer-close:` variants) rather than opening/closing it. But `.drawer-open` set the toggle checkbox to `display: none`, which removes it from the keyboard tab order. As a result, mouse users could still toggle the sidebar size, but **keyboard users could not focus or activate the toggle at all** — exactly the case flagged in #4424 ("the checkbox is hidden but in this case it should not be, because the button toggles the size of drawer").

## Changes

- **`drawer.css`**: drop `@apply hidden` from `.drawer-open > .drawer-toggle`. The checkbox then inherits the base `fixed h-0 w-0 appearance-none opacity-0` — visually hidden but still keyboard-focusable. The always-open panel's visibility/position is unaffected because it is driven entirely by the `~ .drawer-side` sibling rules and `:checked`, none of which depend on the toggle being `display:none`.
- **Docs — icon-only example (`my-drawer-4`)**: add `drawer-button` to the toggle label so the existing focus-forwarding rule (`.drawer-toggle:focus-visible ~ .drawer-content label.drawer-button`) shows the outline on the visible button. Keyboard users can now Tab to it and press Space/Enter to expand/collapse.
- **Docs — always-open example (`my-drawer-3`) + site layout (`+layout.svelte`)**: add `lg:hidden` to the toggle **input**. In the always-open pattern the toggle button is already `lg:hidden`, so without this the now-focusable checkbox would become a phantom (invisible, no focus ring) tab stop at `lg`.
- **Docs — typo**: fix a doubled prefix placeholder `$$$$drawer-button` → `$$drawer-button` in two code samples.

## Before / after

Live demos on Tailwind Play. Because the `.drawer-toggle` checkbox is normally invisible, the demos **reveal it and pin it to the top-left corner** so you can see whether it exists in the DOM while the drawer is open — then press <kbd>Tab</kbd> and watch for a focus ring on the button:

- **Before** (published daisyUI): https://play.tailwindcss.com/hZfNeeqmXa — `.drawer-open > .drawer-toggle` is `display:none`, so the checkbox is removed from the DOM and the tab order. `Tab` never reaches it and the button never gets a focus ring.
- **After** (this PR): https://play.tailwindcss.com/gUNjnPwB0H — the checkbox stays present (still visually hidden in real usage) and keyboard-focusable, so `Tab` reaches it and daisyUI forwards a focus ring to the visible button.

## Compatibility note

This is a small behavior change for the `.drawer-open` modifier. Projects that use the always-open pattern (`lg:drawer-open` + a breakpoint-hidden toggle button) may want to add a matching `lg:hidden` (or similar) to their `drawer-toggle` input to avoid a phantom tab stop at large screens — as done here for the docs examples and the docs site layout. Worth a mention in the changelog.

## Verification

- Rebased onto the latest `master` (resolving a conflict in `+layout.svelte`: kept master's `aria-label` / `<main>` and re-applied the `lg:hidden` on the toggle input).
- `bun test` → 129 pass / 0 fail
- Rebuilt `bun run build` and confirmed the compiled output no longer sets `display:none` on `.drawer-open > .drawer-toggle` (build artifacts are intentionally **not** included in this PR — they're regenerated by the release tooling).
- Ran the docs dev server: the drawer page and site layout compile and render the intended markup; no phantom tab stop on the always-open example, and the icon-only example is now fully keyboard-operable.

Refs #4424

### DIFF
--- a/packages/daisyui/src/components/drawer.css
+++ b/packages/daisyui/src/components/drawer.css
@@ -130,6 +130,8 @@
     }
 
     > .drawer-toggle {
+      @apply hidden;
+
       ~ .drawer-side {
         @apply pointer-events-auto visible sticky block w-auto overscroll-auto opacity-100;
 
--- a/packages/daisyui/src/components/drawer.css
+++ b/packages/daisyui/src/components/drawer.css
@@ -130,8 +130,6 @@
     }
 
     > .drawer-toggle {
-      @apply hidden;
-
       ~ .drawer-side {
         @apply pointer-events-auto visible sticky block w-auto overscroll-auto opacity-100;
 
--- a/packages/docs/src/routes/(routes)/+layout.svelte
+++ b/packages/docs/src/routes/(routes)/+layout.svelte
@@ -162,7 +162,7 @@
     id="drawer"
     type="checkbox"
     autocomplete="off"
-    class={`drawer-toggle ${hasDesktopSidebar ? "lg:hidden" : ""}`}
+    class="drawer-toggle"
     aria-label="Toggle drawer"
     bind:checked
   />
--- a/packages/docs/src/routes/(routes)/+layout.svelte
+++ b/packages/docs/src/routes/(routes)/+layout.svelte
@@ -162,7 +162,7 @@
     id="drawer"
     type="checkbox"
     autocomplete="off"
-    class="drawer-toggle"
+    class={`drawer-toggle ${hasDesktopSidebar ? "lg:hidden" : ""}`}
     aria-label="Toggle drawer"
     bind:checked
   />
--- a/packages/docs/src/routes/(routes)/components/drawer/+page.md
+++ b/packages/docs/src/routes/(routes)/components/drawer/+page.md
@@ -84,7 +84,7 @@ You can check/uncheck the checkbox using JavaScript or by clicking the `label` t
   <input id="my-drawer-1" type="checkbox" class="$$drawer-toggle" />
   <div class="$$drawer-content">
     <!-- Page content here -->
-    <label for="my-drawer-1" class="$$btn $$$$drawer-button">Open drawer</label>
+    <label for="my-drawer-1" class="$$btn $$drawer-button">Open drawer</label>
   </div>
   <div class="$$drawer-side">
     <label for="my-drawer-1" aria-label="close sidebar" class="$$drawer-overlay"></label>
@@ -180,7 +180,7 @@ You can check/uncheck the checkbox using JavaScript or by clicking the `label` t
 #### Sidebar is always visible on large screen, can be toggled on small screen because of lg:drawer-open class
 
 <div class="drawer lg:drawer-open h-56 rounded overflow-hidden">
-  <input id="my-drawer-3" type="checkbox" autocomplete="off" class="drawer-toggle" />
+  <input id="my-drawer-3" type="checkbox" autocomplete="off" class="drawer-toggle lg:hidden" />
   <div class="flex flex-col items-center justify-center drawer-content">
     <label for="my-drawer-3" class="btn drawer-button lg:hidden">Open drawer</label>
   </div>
@@ -195,10 +195,10 @@ You can check/uncheck the checkbox using JavaScript or by clicking the `label` t
 
 ```html
 <div class="$$drawer lg:$$drawer-open">
-  <input id="my-drawer-3" type="checkbox" class="$$drawer-toggle" />
+  <input id="my-drawer-3" type="checkbox" class="$$drawer-toggle lg:hidden" />
   <div class="$$drawer-content flex flex-col items-center justify-center">
     <!-- Page content here -->
-    <label for="my-drawer-3" class="$$btn $$$$drawer-button lg:hidden">
+    <label for="my-drawer-3" class="$$btn $$drawer-button lg:hidden">
       Open drawer
     </label>
   </div>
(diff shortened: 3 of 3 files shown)


packages/daisyui/src/components/drawer.css
```diff
@@ -130,8 +130,6 @@
     }
 
     > .drawer-toggle {
-      @apply hidden;
```
pdanpdan, line 133: Please remove this change - it can be solved in the example with a simple override (the case in the example is an exception, it does not make sense to break usage for everyone)

packages/docs/src/routes/(routes)/components/drawer/+page.md
```diff
@@ -180,7 +180,7 @@ You can check/uncheck the checkbox using JavaScript or by clicking the `label` t
 #### Sidebar is always visible on large screen, can be toggled on small screen because of lg:drawer-open class
 
 <div class="drawer lg:drawer-open h-56 rounded overflow-hidden">
-  <input id="my-drawer-3" type="checkbox" autocomplete="off" class="drawer-toggle" />
+  <input id="my-drawer-3" type="checkbox" autocomplete="off" class="drawer-toggle lg:hidden" />
```
pdanpdan: This is not needed here

packages/docs/src/routes/(routes)/components/drawer/+page.md
```diff
@@ -195,10 +195,10 @@ You can check/uncheck the checkbox using JavaScript or by clicking the `label` t
 
 ```html
 <div class="$$drawer lg:$$drawer-open">
-  <input id="my-drawer-3" type="checkbox" class="$$drawer-toggle" />
+  <input id="my-drawer-3" type="checkbox" class="$$drawer-toggle lg:hidden" />
```
pdanpdan: This is not needed here

packages/docs/src/routes/(routes)/components/drawer/+page.md
pdanpdan: add `inline`

pdanpdan: add `inline`

pdanpdan: add `lg:hidden`
this is the example that has the problem

pdanpdan: add `lg:hidden`
this is the example that has the problem

packages/docs/src/routes/(routes)/+layout.svelte
```diff
@@ -162,7 +162,7 @@
     id="drawer"
     type="checkbox"
     autocomplete="off"
-    class="drawer-toggle"
+    class={`drawer-toggle ${hasDesktopSidebar ? "lg:hidden" : ""}`}
```
pdanpdan: What is the purpose for this? Genuine question

pdanpdan (changes requested):

packages/daisyui/src/components/drawer.css
JamBalaya56562: Done — reverted, `@apply hidden` is back. The component is unchanged for everyone now, and the exception is handled locally in the icon-only example with `inline` (per your comment below). 3d536aba

packages/docs/src/routes/(routes)/components/drawer/+page.md
JamBalaya56562: Added. You are right that this is the one with the problem — this example never uses `drawer-open`, so at `lg` the toggle stayed focusable even though the hamburger next to it is `lg:hidden` and there is no visible control.

JamBalaya56562: Added here too, to keep the snippet in sync with the live example.

packages/docs/src/routes/(routes)/components/drawer/+page.md
JamBalaya56562: Removed — agreed, it is redundant here: this example uses `lg:drawer-open`, so the core `.drawer-open > .drawer-toggle { @apply hidden }` already takes the toggle out at `lg`.

packages/docs/src/routes/(routes)/components/drawer/+page.md
JamBalaya56562: Removed here too.

packages/docs/src/routes/(routes)/components/drawer/+page.md
JamBalaya56562: Added. This is the local override that replaces the core change: `inline` beats the `hidden` from `.drawer-open` (utilities layer wins), while the base `.drawer-toggle` keeps `fixed h-0 w-0 opacity-0` — so the toggle stays invisible but focusable, which this example needs since the button switches the drawer size rather than opening/closing it.

JamBalaya56562: Added here too.

packages/docs/src/routes/(routes)/+layout.svelte
JamBalaya56562: Fair question — it only existed to compensate for the core change in this PR. With `@apply hidden` removed from `.drawer-open > .drawer-toggle`, the docs toggle would have stayed focusable at `lg`, so I hid it here instead. Now that the core is restored, this is redundant (the layout already gets `lg:drawer-open` when `hasDesktopSidebar`), so I reverted it back to `class="drawer-toggle"`.

pdanpdan (approved):

saadeghi:
Thank you

PR #4615, opened 2026-07-05
fix(join): prevent --join-(ss|se|es|ee) propagation to children of .join-item

### LINKED ISSUE #4597: bug: buttons use wrong radius when descendent of join-item

https://play.tailwindcss.com/9BO3aVEu7B

v5.6.5

All browsers

### Describe your issue

With the latest join updates, buttons inside a join-item div now incorrectly reflect the join-item styling.

Observe the differences between the following:
5.5.23 (before) https://play.tailwindcss.com/S2gfZ9sQzX
5.6.1 (after, no border) https://play.tailwindcss.com/VlbYQfaDnz
5.6.5 (after, default border) https://play.tailwindcss.com/9BO3aVEu7B

pdanpdan:
example: https://play.tailwindcss.com/XR1lLHzRSt?file=css

close #4597

### DIFF
--- a/packages/daisyui/src/utilities/join.css
+++ b/packages/daisyui/src/utilities/join.css
@@ -42,6 +42,15 @@
 }
 
 .join-item {
+  @layer daisyui.l1.l2.l3.l4 {
+    > * {
+      --join-ss: initial;
+      --join-se: initial;
+      --join-es: initial;
+      --join-ee: initial;
+    }
+  }
+
   border-style: solid;
   border-width: var(--border, 1px);
   border-start-start-radius: var(--join-ss);


pdanpdan:
There is a DX problem with setting the radius for the corners - as it is defined now setting the `--join-(ss|se|es|ee)` on `.join` has no effect because the radius is always computed based on `--radius-field`. They must be defined on the children.

I suggest defining some computed values based on `--join-(ss|se|es|ee)` (`--join-i-(ss|se|es|ee)`?), and defaulting `--join-(ss|se|es|ee)` to `var(--radius-field)`.
Then the internal computed `--join-i-(ss|se|es|ee)` values should be used in button/fileinput/input/select/join

@saadeghi please let me know if it soulds good to you and if you want the changes in this PR or in a new one.

saadeghi:
Thanks.

> There is a DX problem with setting the radius for the corners

It can be a new PR, but can give a Tailwind Play example so I can see in which situation it would make an improvement?

pdanpdan:
I'll create an issue for it

Issue #4591, opened 2026-06-27
bug: button with btn-outline btn-neutral btn-active together are illegible


kuchta:
### Reproduction URL (Required)

https://play.tailwindcss.com/8VYfzZ7imo

### What version of daisyUI are you using?

5.6.3

### Which browsers are you seeing the problem on?

Chrome

### Describe your issue

if I use all three modifiers (btn-outline btn-neutral btn-active) button text is illegible basically under any theme

saadeghi:
How do you expect it to look like?

pdanpdan:
I think there is a problem with the fg color applied by .btn-active
~https://play.tailwindcss.com/IQfmVG6wiE~
https://play.tailwindcss.com/rSE8vDTtsC
On all colors and styles except filled

kuchta:
I don't care @saadeghi, just not like these (in first 5 themes)

<img width="59" height="36" alt="Image" src="https://github.com/user-attachments/assets/0a80b89b-0bbd-4c9c-a4c2-f2c61c47194b" />

<img width="59" height="37" alt="Image" src="https://github.com/user-attachments/assets/12543c28-4fb2-4dad-b2df-9b956691b583" />

<img width="60" height="36" alt="Image" src="https://github.com/user-attachments/assets/be20a6a0-b3a1-4e01-a81c-96ac5d372604" />

<img width="67" height="34" alt="Image" src="https://github.com/user-attachments/assets/e5b1f129-2de8-49d7-a2ac-dd0cc5277ee8" />

<img width="65" height="39" alt="Image" src="https://github.com/user-attachments/assets/c16a70c7-1f31-4e53-9edd-7253e00e4b5b" />

saadeghi:
> I don't care [@saadeghi](https://github.com/saadeghi), just not like these (in first 5 themes)

I understand, I'm asking what's your expected behavior when you use `btn-outline btn-neutral btn-active` together?  
Do you want it to look like `btn-neutral btn-active`?

kuchta:
In one [application](https://kuchta.github.io/radixverse) if you open settings, you will find a bunch of radix buttons. They just looked best with btn-outline and btn-neutral.

Previously (until 5.6.0 inclusive) it looked like this:

<img width="355" height="252" alt="Image" src="https://github.com/user-attachments/assets/92819f94-c82c-40a2-94fd-14773d296a0b" />

After 5.6.0:

[...]

PR #4619, opened 2026-07-08
fix(btn): move btn-active in layer l2, interactive styles in l1, and link overrides in daisyui layer

pdanpdan:
reorder specificity so that:
- base btn is in l3
- btn-active is in l2
- interactivity is in l1
- link overrides in daisyui layer
- disabled overrides in daisyui layer

I'm not sure about the last 2 (link and disabled) - the should be safe but another person checking would help

example: https://play.tailwindcss.com/aRHZQWxcH7

ref #4591

### DIFF
--- a/packages/daisyui/src/components/button.css
+++ b/packages/daisyui/src/components/button.css
@@ -65,7 +65,7 @@
     }
   }
 
-  @layer daisyui.l1.l2 {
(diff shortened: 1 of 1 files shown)


saadeghi:
Thanks

PR #4636, opened 2026-07-21
fix(select,input,textarea,rating): restore -xs/-sm size modifiers in the packed CSS

### LINKED ISSUE #4631: bug: 5.6以上版本，-sm不生效

见下方内容

v5.6.x

_No response_

### Describe your issue

```html
<!DOCTYPE html>
<html>

<head>
  <link href="https://cdn.jsdelivr.net/npm/daisyui@5" rel="stylesheet" type="text/css" />
  <script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>
</head>

<body>

  <select class="select select-xs">
    <option disabled selected>Xsmall</option>
    <option>Xsmall Apple</option>
    <option>Xsmall Orange</option>
    <option>Xsmall Tomato</option>
  </select>
  <select class="select select-sm">
    <option disabled selected>Small</option>
    <option>Small Apple</option>
    <option>Small Orange</option>
    <option>Small Tomato</option>
  </select>
  <select class="select select-md">
    <option disabled selected>Medium</option>
    <option>Medium Apple</option>
    <option>Medium Orange</option>
    <option>Medium Tomato</option>
  </select>
  <select class="select select-lg">
    <option disabled selected>Large</option>
    <option>Large Apple</option>
    <option>Large Orange</option>
    <option>Large Tomato</option>
  </select>
  <select class="select select-xl">
    <option disabled selected>Xlarge</option>
    <option>Xlarge Apple</option>
    <option>Xlarge Orange</option>
    <option>Xlarge Tomato</option>
  </select>

</body>

</html>
```

Image

JamBalaya56562:
## Problem

Closes #4631 — Ref discussions/4620

Since v5.6, `select-xs` / `select-sm` (and the same sizes on `input`, `textarea`, `rating`) render identical to `-md` **when daisyUI is consumed as the packed CSS file** (the CDN / `daisyui.css`).

v5.6 moved sizing to multipliers (`--size: calc(var(--size-field) * var(--sl-size-mul))`) and put the md defaults into grouped selectors like:

```css
.select-xs { --sl-size-mul: 6; … }
.select-sm { --sl-size-mul: 8; … }
.select, .select-md { --sl-size-mul: 10; … }   /* ← after -xs/-sm in the source */
```

In the packed file these all land in the same layer (`daisyui.l1.l2`) with the same `(0,1,0)` specificity, so for `<select class="select select-xs">` the later defaults rule wins by **source order** and `-xs`/`-sm` collapse to md. `-lg`/`-xl` sit after the defaults, which is why only the two smaller sizes broke — exactly what @pdanpdan diagnosed in #4620: *"the order how they are generated — now the select and select-md are generated after select-xs and select-sm"*.

### Why it "worked" on daisyui.com but not on the CDN

Both observations in #4620 were correct: the **Tailwind plugin path is unaffected** because Tailwind splits the grouped selector and emits the base `.select` (with the defaults) early and each size modifier later — so the docs site looks fine. The **packed file keeps source order**, so CDN users hit the bug (reporter's live repro: https://cristhiancustodio.github.io/daisy-example/).

## Fix

Restore the pre-5.6 arrangement: the md defaults live on the **base component block** (`@layer daisyui.l1.l2.l3` — deeper layer = weaker in daisyUI's layer design), so any size modifier (`daisyui.l1.l2`) wins **regardless of source order**, and `.component, .component-md` is ungrouped into a standalone `.component-md`.

- `select.css`, `input.css`, `textarea.css`: defaults (incl. the `.floating-label:has(&)` vars) moved to the base block + ungrouped.
- `rating.css`: the base already had the default `--size`; it only needed ungrouping (the `.rating,` prefix was re-asserting md after `-xs`/`-sm`).

**https://play.tailwindcss.com/CaPMTXIZR7**

Play compiles through the plugin path where the bug can't reproduce, so the demo replays the packed-CSS cascade literally on real daisyUI selects: "Before" orders the same-specificity rules like the packed file (`-xs` first, defaults after → renders md-size); "After" puts the defaults in a weaker layer like this PR (→ renders xs-size).

## Verification (packed CSS in Chromium)

| metric | before (xs / sm / md / lg / xl) | after |
|---|---|---|
| select height | 40 / 40 / 40 / 48 / 56 px | **24 / 32 / 40 / 48 / 56** |
| input height | 40 / 40 / 40 / 48 / 56 px | **24 / 32 / 40 / 48 / 56** |
| textarea font-size | 14 / 14 / 14 / 18 / 22 px | **11 / 12 / 14 / 18 / 22** |
| rating item size | 24 / 24 / 24 / 28 / 32 px | **16 / 20 / 24 / 28 / 32** |

The Tailwind plugin path still yields 5 distinct sizes for all four components (unchanged — no regression). Test suite: 177 pass / 0 fail.

### DIFF
--- a/packages/daisyui/src/components/input.css
+++ b/packages/daisyui/src/components/input.css
@@ -10,11 +10,6 @@
     --font-size-min: 0.875rem;
     --spin-my: -3;
 
-    .floating-label:has(&) {
-      --top-mul: 5;
-      --font-size: 0.875rem;
-    }
-
     cursor: text;
     width: clamp(3rem, 20rem, 100%);
     height: var(--size);
--- a/packages/daisyui/src/components/input.css
+++ b/packages/daisyui/src/components/input.css
@@ -4,6 +4,16 @@
 
     --size: calc(var(--size-field, 0.25rem) * var(--in-size-mul));
     --input-color: color-mix(in oklab, var(--color-base-content) 20%, #0000);
+    /* md defaults live on the base (deeper layer = weaker) so size modifiers
+       win regardless of source order in the packed CSS */
+    --in-size-mul: 10;
+    --font-size-min: 0.875rem;
+    --spin-my: -3;
+
+    .floating-label:has(&) {
+      --top-mul: 5;
+      --font-size: 0.875rem;
+    }
 
     cursor: text;
     width: clamp(3rem, 20rem, 100%);
@@ -242,7 +252,6 @@
   }
 }
 
-.input,
 .input-md {
   @layer daisyui.l1.l2 {
     --in-size-mul: 10;
--- a/packages/daisyui/src/components/input.css
+++ b/packages/daisyui/src/components/input.css
@@ -2,18 +2,13 @@
   @layer daisyui.l1.l2.l3 {
     @apply bg-base-100 relative inline-flex shrink appearance-none items-center gap-2 px-3 align-middle whitespace-nowrap;
 
-    --size: calc(var(--size-field, 0.25rem) * var(--in-size-mul));
+    --size: calc(var(--size-field, 0.25rem) * var(--in-size-mul, 10));
     --input-color: color-mix(in oklab, var(--color-base-content) 20%, #0000);
-    /* md defaults live on the base (deeper layer = weaker) so size modifiers
-       win regardless of source order in the packed CSS */
-    --in-size-mul: 10;
-    --font-size-min: 0.875rem;
-    --spin-my: -3;
 
     cursor: text;
     width: clamp(3rem, 20rem, 100%);
     height: var(--size);
-    font-size: max(var(--font-size, 0rem), var(--font-size-min));
+    font-size: max(var(--font-size, 0rem), var(--font-size-min, 0.875rem));
     touch-action: manipulation;
     border-start-start-radius: var(--join-ss, var(--radius-field));
     border-start-end-radius: var(--join-se, var(--radius-field));
@@ -69,7 +64,7 @@
       }
 
       &::-webkit-inner-spin-button {
-        margin-block: calc(var(--spacing) * var(--spin-my));
+        margin-block: calc(var(--spacing) * var(--spin-my, -3));
       }
 
       &::-webkit-calendar-picker-indicator {
--- a/packages/daisyui/src/components/rating.css
+++ b/packages/daisyui/src/components/rating.css
@@ -63,7 +63,6 @@
   }
 }
 
-.rating,
 .rating-md {
   @layer daisyui.l1.l2 {
     --size: var(--size-selector, 0.25rem) * 6;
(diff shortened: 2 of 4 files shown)


pdanpdan:
I think a cheaper solution would be to put the combined `.component, .component-md` first in the list of sizes (before xs)

JamBalaya56562:
Thanks @pdanpdan — you're right that reordering the combined `.select, .select-md` before `-xs` would also fix it, and it's a smaller diff. But I'd prefer to keep the layer-based approach in this PR, for two reasons:

1. **Consistency with the rest of daisyUI.** Every other component already puts its base block (with the md defaults) in `@layer daisyui.l1.l2.l3` and all modifiers — colors *and* sizes — in `@layer daisyui.l1.l2`. `input`, `textarea`, `rating` and `select` all follow this. The v5.6 regression was essentially these size defaults ending up in the modifier layer (`l2`) instead of the base layer (`l3`) like everywhere else. Moving them back to the base restores that convention rather than making these four components a special case.

2. **It removes the source-order dependency entirely.** Reordering keeps the fix relying on the defaults rule staying ahead of `-xs`/`-sm` at the same specificity — which is exactly the fragility that caused this bug in the first place. With the defaults in the weaker `l3` sub-layer, any size modifier in `l2` wins regardless of source order, so a future reordering can't reintroduce the regression.

Since the base block already lives in `l3`, putting the md defaults there is the natural home for them. Happy to reconsider if you see a downside to the layer approach though!

pdanpdan:
I checked again the diff.

For rating.css it's clear.

For input.css:
- we remove line 245 (`.input,`)
- then use defaults for `--in-size-mul`, `--font-size-min`, `--spin-my` (`--top-mul` and `--font-size` already have defaults in label.css)

For select.css:
- we remove line 301 (`.select,`)
- then use defaults for `--sl-size-mul`, `--font-size-min`, `--option-px` (`--top-mul` and `--font-size` already have defaults in label.css)

For textarea.css:
- we remove line 184 (`.textarea,`)
- then use defaults for `--font-size-min` (`--top-mul` and `--font-size` already have defaults in label.css)

I would like to avoid adding all the variables.

JamBalaya56562:
Good catch @pdanpdan — you're right, `--top-mul` and `--font-size` already fall back to their md values in `label.css`, so the `.floating-label` block on the base was redundant. Dropped it from all three; only the vars consumed without a fallback (`--*-size-mul`, `--font-size-min`, `--option-px`, `--spin-my`) stay on the base now. Pushed.

pdanpdan:
I was thinking of also adding the default values for the rest of the variables - they are only used in one place for each component

JamBalaya56562:
Done 👍 Each of those (`--sl-size-mul`/`--in-size-mul`, `--font-size-min`, `--option-px`, `--spin-my`) is consumed in exactly one place, so I moved the md default into the `var(..., default)` fallback there and dropped the base-block declarations — same pattern `label.css` uses for `--top-mul`/`--font-size`. The base no longer declares any of them, so there's nothing left to lose a source-order battle.

Rebuilt & verified the packed CSS: base fallback resolves to md when no size modifier is set, and heights are still xs/sm/md/lg/xl = 24/32/40/48/56px.

pdanpdan (approved):

saadeghi:
Thank you

PR #4638, opened 2026-07-22
Fix(modal): Parent modal styles leak into nested native dialog modal-box

IdrisCelik:
bug: Parent modal styles leak into nested native dialog modal-box

### DIFF
--- a/packages/daisyui/src/components/modal.css
+++ b/packages/daisyui/src/components/modal.css
@@ -43,7 +43,7 @@
         opacity 0.1s ease-out;
       background-color: oklch(0% 0 0/ 0.4);
 
-      .modal-box {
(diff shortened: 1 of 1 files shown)


pdanpdan:
I think it's pretty safe to assume `.modal > .modal-box`
But please remove the leading `&` - it's not needed

example: https://play.tailwindcss.com/2F1Uyh3252

IdrisCelik:
Good point, thanks! I used & to make the parent selector explicit, but DaisyUI’s nested component object already resolves > .modal-box relative to .modal, so it’s redundant. I’ve removed it.

pdanpdan:
Don't forget to push

saadeghi:
Thank you 🌼

Issue #3722, opened 2025-03-25
bug: wrong outline in join element


vchirikov:
### Reproduction URL (Required)

https://daisyui.com/components/join/

### What version of daisyUI are you using?

v5.0.9

### Which browsers are you seeing the problem on?

All browsers

### Describe your issue

Hi, first of all, thanks for your work!
I found a regression between v4 and v5, just use tab/focus to travers between join items
v5:

![Image](https://github.com/user-attachments/assets/dbd55323-cfa3-4a09-8395-53e67ed5b47c)

v4:
![Image](https://github.com/user-attachments/assets/a0bd8d3b-063b-4778-b6cc-c88a0e127e4c)

Same with first item, only last join-item works as expected

pdanpdan:
Add a `focus-visible:z-1` on `join-items`

vchirikov:
@pdanpdan big thanks, z-index helps :)

saadeghi:
I should add isolation to input and select as well. So if they are next to each other in any way, (not necessarily in a `join`), their outline won't hide under another.

pdanpdan:
I tried it but it doesn't seem to work as expected (https://play.tailwindcss.com/MJcWoKqhhZ?file=css).
I'll check again a little bit later.

saadeghi:
> I _should_ add

In next patch tomorrow 👍

pdanpdan:
I really meant it when I said `isolate` is not working for input and select :)

saadeghi:
@pdanpdan Yeah… That's unfortunate.  
To fix that I will add `z-1` when input is focused.  
And for disabled buttons, I will use [your suggestion about `z-index: -1`](https://github.com/saadeghi/daisyui/issues/3722#issuecomment-2754560834)

saadeghi:
for disabled buttons, `z-index: -1` was not a good idea. it was causing issues when other elements in the background was having z-index, and it was hiding the disabled button behind them. Instead I removed the minus margin of `join-item` for disabled button so it all looks good now.

<img width="272" alt="Image" src="https://github.com/user-attachments/assets/e52980c9-b2e6-4f25-8fc7-7f250d59c397" />

<img width="817" alt="Image" src="https://github.com/user-attachments/assets/eceb8faa-04a9-4314-aa15-b9f360a58584" />

<img width="447" alt="Image" src="https://github.com/user-attachments/assets/4cf2ca68-ea31-4ccf-9764-61ec84d6c363" />

[...]

PR #4642, opened 2026-07-26
fix(join): use z-index instead of isolation so tooltips aren't trapped

### LINKED ISSUE #4618: bug: Possible conflict between join/join-item and tooltip, when it comes to z-index/display priority

https://play.tailwindcss.com/Bei4oumtKN

5.6.14 & 5.6.15 (5.5.x were perfectly OK) (Maybe the bug appears for any 5.6.x, I didn't test)

Edge

### Describe your issue

Link to sandbox:
[https://play.tailwindcss.com/Bei4oumtKN](https://play.tailwindcss.com/Bei4oumtKN)

Hello,

I have an issue with latest version 5.6, having a conflict when displaying tooltips. IMO, it's because of the join/join-item, because, when I remove them, it becomes functional.

Here is a demo GIF of the actual issue:
Image

Thank you! :)

JamBalaya56562:
## Problem

Closes #4618

A tooltip on a join item is painted **under the next join item**.

`join.css` puts the hovered item in a stacking context with `isolation: isolate`. That creates a stacking context but leaves `z-index: auto`, so it only lifts the item to the **`z-index: 0`** layer — and any *later* sibling in that same layer wins on DOM order. In a join that happens constantly, because the later item is typically positioned too:

- `.tooltip` → `position: relative` (`tooltip.css:3`)
- `.input` / `.select` / `.dropdown` → `position: relative`
- `.btn-active` / `:checked` / `:focus-visible` → `isolation: isolate` (`button.css:109,116,129`)

Measured with `elementsFromPoint` against the built CSS:

| hovered item | next sibling | topmost |
|---|---|---|
| *(nothing)* | plain `.btn` | next ← this was #4320 |
| `isolation: isolate` | plain `.btn` | hovered ← the #4320 fix, works |
| `isolation: isolate` | **positioned** (tooltip/input/select/dropdown) | next ← **#4618** |
| `z-index: 1` | positioned | hovered ✅ |

**This also means `isolation` was never a complete fix for #4320**: today, hovering a `.btn.join-item` followed by an `.input`/`.select`/`.tooltip` still fails to highlight the shared border. This PR widens that fix rather than narrowing it.

## Fix

```diff
     > :where(:focus, :has(:focus)) {
-      @apply z-1;
+      @apply z-2;
     }
 
     @media (hover: hover) {
       > :where(.btn:hover, :has(.btn:hover)) {
-        @apply isolate;
+        @apply z-1;
       }
     }
```

Flex items honor `z-index` even while `position: static` (Flexbox spec §4) — which is exactly why the `:focus` rule three lines above already works without `.btn` having a `position`. `z-index: 1` puts the hovered item in the positive-z layer, above positioned siblings.

**Focus goes to `z-2`, and that part isn't cosmetic**: with both rules at `z-1`, an item hovered *after* another item was focused wins on DOM order and clips the focused item's 2px outline. Verified in both orderings.

**`isolation` isn't kept alongside `z-index`**: `z-index: 1` on a flex item creates a stacking context by itself, so it would be dead weight.

**`position: relative` isn't added either**: it isn't needed (paint order is identical without it), and because the rule is conditional it would create a containing block *only while hovered* — making absolutely-positioned descendants jump on hover.

**https://play.tailwindcss.com/TUbFzhmShE**

Play loads the released daisyUI, so the "After" join carries a small shim reproducing this patch. Both rows pin the hovered state so the difference is visible without a pointer: Before the tooltip is completely hidden behind the next button, After it paints on top.

## Verification

Chromium, `document.elementsFromPoint()` at the overlap point, against the real built CSS (`--blink-settings=...HoverType=2` so `@media (hover: hover)` matches in headless):

| case | before | after |
|---|---|---|
| **#4618** — hovered tooltip item vs next item | next item ❌ | **tooltip** ✅ |
| **#4320** — plain buttons, shared border | hovered | hovered (unchanged) ✅ |
| earlier focus, later hover | focus wins | focus wins (unchanged) ✅ |
| earlier hover, later focus | focus wins | focus wins (unchanged) ✅ |
| **#4320 widened** — hovered `.btn` followed by `.input` | `.input` ❌ | **hovered `.btn`** ✅ |

Computed style on the hovered item flips from `isolation: isolate; z-index: auto` to `isolation: auto; z-index: 1`. Screenshots confirm it visually. Build + test suite pass (one pre-existing, unrelated docs-translation timeout fails with and without this change).

## Note — related, deliberately out of scope

`button.css:109/116/129` (`:checked`, `:focus-visible`, `.btn-active`) set `isolation: isolate` and cause the **same** trapping *outside* of `.join` — two adjacent `.btn-active`s, the first with a `.tooltip-right`, hides the tooltip under the second. It's a different component with no shared code path, those rules exist to keep the focus outline from being clipped, and `.btn` has no flex container guaranteeing `z-index` would work, so it needs its own `position` decision. Happy to file it separately.

The residual case after this PR: a join item that is `.btn-active`/`:checked` but neither hovered nor focused still traps its own `tooltip-open`, for that same reason.

### DIFF
--- a/packages/daisyui/src/utilities/join.css
+++ b/packages/daisyui/src/utilities/join.css
@@ -9,12 +9,12 @@
 
   @scope (&) {
     > :where(:focus, :has(:focus)) {
-      @apply z-1;
+      @apply z-2;
     }
 
     @media (hover: hover) {
       > :where(.btn:hover, :has(.btn:hover)) {
-        @apply isolate;
+        @apply z-1;
       }
     }
 


pdanpdan:
I think it looks good: https://play.tailwindcss.com/DRUq8BgNrH

saadeghi:
Thanks!

pdanpdan:
@JamBalaya56562 can you please check if isolation is still needed in the components where it was added because of `.join`?
I'll do some research myself

JamBalaya56562:
@pdanpdan sure — I'm looking into it now, so we don't duplicate effort.

Scope I'm checking, with measurements rather than reasoning:

- `input.css:88`, `select.css:75` — the two `isolation`s that #4320 (`5a37a131`) deliberately kept when it removed their `z-index: 1`
- `textarea.css:36`, `fileinput.css:61`, `otp.css:137` — same `:focus`/`:focus-within` shape, not touched by #4320
- `button.css:109/116/129` (`:checked`, `:focus-visible`, `.btn-active`) — the ones I flagged in the PR body as trapping tooltips outside of `.join`

For each I'm ablating the declaration against the real built CSS and checking with `elementsFromPoint` whether anything actually changes: focused item vs both neighbours inside a join, the #4320 shared-border case, and the same components outside a join (where `.input` is already `position: relative`, unlike `.btn`).

I'll post the per-component results with a Play link. If some turn out to be redundant I'll open a PR; if they're still doing work, I'll say so.

JamBalaya56562:
@pdanpdan here's what I found. Short version: **6 of the 8 can go with zero measured difference; the 2 on `.btn` are genuinely working, but only inside `.join` — and outside it they're actively harmful.**

Play with all three cases side by side: **https://play.tailwindcss.com/A0goBWX7Bf**

## Method

Ablated each declaration against the real built `daisyui.css` in Chromium and compared paint order with `elementsFromPoint` plus pixel sampling of the focus ring / seam / tooltip bubble. Then physically deleted the "redundant" ones, rebuilt, and re-ran everything.

Two structural facts drive most of the results:

- `.input` and `.select` are **already `position: relative`** (`input.css:3`, `select.css:3`), so they're in the same paint step as a positioned sibling with or without `isolation` — it cannot change sibling order for them.
- `.otp` sets `clip-path: inset(...)` unconditionally (`otp.css:7`), and **`clip-path` already creates a stacking context** (verified in isolation), so `otp.css:137` is structurally dead.

Also worth noting from the history: #3722 (`78bf76dd`, 2025-04) added `z-index: 1` **on top of** the existing `isolation` precisely because isolation alone didn't fix the seam. #4320 then moved that `z-index` to `.join`. So `isolation` was never the thing solving it.

## Results

| declaration | verdict | evidence |
|---|---|---|
| `input.css:88` | **redundant** | every condition byte-identical with/without; no-op by construction (already `relative`) |
| `select.css:75` | **redundant** | same |
| `textarea.css:36` | **redundant** | `.join`'s `z-2` covers it; only a hand-rolled negative-margin overlap outside `.join` would notice |
| `fileinput.css:61` | **redundant** | same |
| `otp.css:137` | **redundant** | `clip-path` already isolates; screenshot SHA identical |
| `button.css:116` (`:focus-visible`) | **redundant** | no measured difference — and in the case where you'd *want* it (next sibling is `.btn-active`) it doesn't help anyway |
| `button.css:109` (`:checked`) | **still needed, but only in `.join`** | removing it hands the shared seam to the neighbour (measured: seam pixels go `61,38,199` → `232,232,232`) |
| `button.css:129` (`.btn-active`) | **same** | seam `59,37,193` → `232,232,232` |

Deleting the six redundant ones and rebuilding: **0 differences across 8 scenarios** (join of 3× input / select / textarea / file-input / btn, a mixed btn|input|btn join, the #4618 tooltip-in-join case, and an OTP screenshot hash).

## The `.btn` two

They do real work, but the stacking context they create is what buries tooltips (section B in the Play — the bubble reads "OLTIP" because it's behind the next button). Replacing them with a join-scoped `z-index` keeps the seam and frees the tooltip:

```css
/* button.css: drop isolation from :checked and .btn-active */
/* join.css, inside @scope (&): */
> :where(.btn-active, :checked) {
  @apply z-1;
}
```

Measured: seam returns to `59,37,193`, tooltip goes from buried to on top. It also fixes a smaller existing bug — right now a `:focus-visible` button next to a `.btn-active` sibling has its focus ring painted over (`232,232,232` instead of `24,24,27`); with the change the ring survives.

One caveat I should state plainly: `z-index` only re-creates the stacking inside `.join`, so anyone overlapping `.btn-active` / `:checked` with their own negative margins outside a join would see a change. daisyUI itself only overlaps siblings in `.join` (`.filter` uses a positive gap), so I think that's acceptable — but it is a behavior change, unlike the other six.

## Proposal

Two PRs, so the risk-free part isn't held up by the judgement call:

1. **Pure removal** — the six redundant declarations. Zero measured difference, #4320 and #4618 both still hold.
2. **`.btn` stacking** — drop the two `isolation`s, add the join-scoped `z-index`. Fixes the tooltip trap and the focus-ring case.

Happy to open #1 right away and leave #2 for you/@saadeghi to weigh in on first — or both, whichever you prefer.

pdanpdan:
Can you please open 3 PRs?
- one with `input` and `select` because I think they are pretty clear
- one with the rest except `btn`
- one for btn

If you can please also add a play example for each PR with a toggle on top like:
```html
<label class="label apply-change m-4">
  <input type="checkbox" class="checkbox" checked />
  Remove isolation
</label>

<div class="m-4">
  ... examples here
</div>
```

and in the CSS section use
```css
.apply-change:has(:checked) + div {
  rule {
    &.that_needs_to_change {
      isolation: auto;
    }
  }
  ...
}
```

So that it is easy to check the result with current published daisyUI and with the proposed changes.

saadeghi:
A few hours ago I added a GitHub Action that automatically generates and comments Tailwind Play before/after links to new PRs. Hope it makes it easier to compare the changes visually.

JamBalaya56562:
Follow-up to my audit above — **I have to walk part of it back.**

My original measurements only overlapped the components with *positioned* siblings, so `isolation` looked like a no-op everywhere. Re-tested against a **plain static** sibling, only three of the eight declarations are actually redundant:

- **no-op**: `.input:focus`, `.select:focus` (already `position: relative`), `.otp:focus-within` (already isolated by its `clip-path`)
- **load-bearing**: `.textarea:focus`, `.file-input:focus`, `.btn:focus-visible` / `:has(:focus-visible)`, `.btn-active`, `.btn:checked`

`isolation: isolate` turns a non-positioned element into a stacking context, which paints in step 8 of the CSS 2.1 painting order — after plain in-flow and inline-level siblings. So it keeps a focused/active control **and its outline** above adjacent content. That is not the `.join` seam (that's `z-index`'s job, per #3722 → #4320 → #4642), but it is a real job.

Demo — toggle the checkbox and watch each component drop behind the red block: https://play.tailwindcss.com/LEyZbb15eR

Result for the three PRs you asked for: #4656 (`input`/`select`) stands, #4657 is narrowed to `otp` alone, and I am **not** opening the `btn` one. Full details in https://github.com/saadeghi/daisyui/pull/4657#issuecomment-5157036294.

PR #4652, opened 2026-08-01
fix: allow user override of content in calendar and menu

pdanpdan:


pdanpdan:
I reverted the formatting changes in menu.
I think a format on repo is better

PR #4653, opened 2026-08-01
fix: move tooltip default position in l3 in main definition

### LINKED ISSUE #4647: bug: tooltip-bottom broken

https://daisyui.com/components/tooltip/#bottom

5.7.8

Chrome

### Describe your issue

using style `tooltip tooltip-bottom` seems to just render as if it was `tooltip tooltip-top`

pdanpdan:
example: https://play.tailwindcss.com/iT2qxCVOb1

close #4647


saadeghi:
Thanks!

PR #4654, opened 2026-08-01
fix: force LTR for special case input types

### LINKED ISSUE #4644: bug: `email` Input RTL Placeholder Is Misplaced

https://play.tailwindcss.com/itfMld73Tj

5.7.4

Chrome

### Expected Behavior 
The placeholder is aligned according to the directionality rule. Same like it does for native non daisy inputs.

pdanpdan:
According to https://simplelocalize.io/blog/posts/rtl-design-guide-developers/#forms-in-rtl some types of input should be LTR always (email, tel, url, number).

Changes:
- defaults text align to start for all except number (where it is end)
- apply LTR direction for all special types
- fix join border radius for special types

example: https://play.tailwindcss.com/xYeuSceT4C

close #4644

### DIFF
--- a/packages/daisyui/src/components/input.css
+++ b/packages/daisyui/src/components/input.css
@@ -50,10 +50,6 @@
     input {
       @apply relative inline-flex text-start;
 
-      &[type="number"] {
-        @apply text-end;
-      }
-
       &[type="url"],
       &[type="tel"],
       &[type="email"],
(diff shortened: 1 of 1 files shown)


saadeghi (changes requested):
Thanks 👍  
Similar to `type="tel"`, the `type="number"` inputs must be always LTR.
In RTL languages, numbers are still LTR, and aligned to left (must not have `text-align: end;`)

pdanpdan:
I was thinking that it's more common to right-align numbers (in LTR orientation).
I'll remove that.

PR #4657, opened 2026-08-02
refactor(otp): remove redundant isolation on focus-within

JamBalaya56562:
> **Corrected.** The first version of this PR also removed `isolation` from `textarea`, `file-input` and `button`. That was wrong — I re-tested against a plain *static* overlapping sibling and those three declarations **do** change paint order. Only `.otp` is genuinely a no-op, so this PR is now one line. Details and evidence in the comment below.

Part of the 3-PR split @pdanpdan asked for in https://github.com/saadeghi/daisyui/pull/4642#issuecomment-5156748718.

- PR 1 (`input` + `select`): #4656 — still valid, both are already `position: relative`
- **this PR**: `otp` only
- there is no third PR: `.btn`'s two `isolation` declarations turned out to be load-bearing (see below)

## The change

```diff
     &:focus-within {
-      isolation: isolate;
       --input-color: var(--color-base-content);
```

`.otp` already establishes a stacking context via
[`clip-path: inset(-3.5px 3.5px -3.5px -3.5px)`](https://github.com/saadeghi/daisyui/blob/master/packages/daisyui/src/components/otp.css#L7)
on the base class, so the `isolation: isolate` added on `:focus-within` can never
have an effect — the element is already isolated in every state.

## Play demo

**https://play.tailwindcss.com/KWt5JfPqu2**

Play ships released daisyUI, which still has the declaration, so the checkbox at
the top neutralises exactly the line this PR removes:

```css
.apply-change:has(:checked) + div {
  .otp { &:focus-within { isolation: auto; } }
}
```

Click into each field and toggle — plain `.otp`, `.otp` under an open tooltip,
`.otp` overlapped by a static sibling, and `.otp-joined` are all unaffected.

## Verification

Built `master` and this branch and compared the two `daisyui.css` files headlessly
(Chromium, `hover: hover` forced, 3 samples per side, transitions and caret blink
disabled to remove flake). Screenshots **and** `document.elementsFromPoint()` at
both seams are identical for:

- `.otp` focused, standalone
- `.otp` focused, overlapped by an open `.tooltip`
- `.otp` focused, overlapped by a plain static `<div>` with a negative margin
- `.otp-joined`

The two builds differ by exactly `-108` bytes = 1 declaration × 6 responsive copies × 18 bytes,
and stripping `isolation: isolate` from both makes them byte-identical, so nothing else moved.

### DIFF
--- a/packages/daisyui/src/components/otp.css
+++ b/packages/daisyui/src/components/otp.css
@@ -134,7 +134,6 @@
     }
 
     &:focus-within {
-      isolation: isolate;
       --input-color: var(--color-base-content);
       > span {
         box-shadow: 0 1px color-mix(in oklab, var(--input-color) calc(var(--depth) * 10%), #0000);


pdanpdan:
Fixed example :)
https://play.tailwindcss.com/m2ESaGNhcv

JamBalaya56562:
@pdanpdan I have to correct myself on this one, and it also answers the third PR you asked for.

**I was wrong that these declarations are redundant.** My first round of fixtures only overlapped the components with *positioned* siblings (`.tooltip`, `.join` items that already carry a `z-index`). Against those, `isolation` genuinely makes no difference — which is why everything looked identical. But `isolation: isolate` makes a **non-positioned** element a stacking context, and a stacking context with `z-index: auto` paints in step 8 of the CSS 2.1 painting order, i.e. *after* plain in-flow and inline-level siblings. So against a **plain static sibling** it very much does something.

Measured with `document.elementsFromPoint()` at the overlap (component vs. a bare `<div>` with no `position` and no `z-index`):

| declaration | with `isolation` | removed | verdict |
| --- | --- | --- | --- |
| `.input:focus` | `input > div` | `input > div` | **no-op** (already `position: relative`) |
| `.select:focus` | `select > div` | `select > div` | **no-op** (already `position: relative`) |
| `.otp:focus-within` | unchanged | unchanged | **no-op** (already isolated by `clip-path`) |
| `.textarea:focus` | `textarea > div` | `div > textarea` | **load-bearing** |
| `.file-input:focus` | `file-input > div` | `div > file-input` | **load-bearing** |
| `.btn:focus-visible` / `:has(:focus-visible)` | `btn > div` | `div > btn` | **load-bearing** |
| `.btn-active` | `btn > div` | `div > btn` | **load-bearing** |
| `.btn:checked` | `btn > div` | `div > btn` | **load-bearing** |

Play demo of the load-bearing cases — toggle the checkbox and watch the component drop behind the red block, focus ring included:
**https://play.tailwindcss.com/LEyZbb15eR**

For `.btn:checked` inside a `.join` the effect is subtler but visible: the checked button's shared border currently paints over its neighbour's, and without `isolation` the *next* sibling wins instead. Pixel diff of that seam, `master` vs. `isolation` removed — 109 differing pixels, max channel delta 194/255, in a 3px-wide strip exactly on the seam.

So the practical effect of these declarations is **not** the `.join` seam (that is `z-index`'s job, as #3722 → #4320 → #4642 showed) — it is keeping a focused/active control and its outline above adjacent content in ordinary layouts. That is worth keeping.

**Where that leaves the three PRs:**

- #4656 (`input` + `select`) — still correct, both are already `position: relative` so `isolation` provably cannot change their paint order.
- this PR — narrowed to `otp` alone, the only remaining true no-op.
- the third PR (`btn`) — **I am not opening it.** Removing `isolation` from `.btn-active` / `.btn:checked` is a real regression, and replacing it with a `.join`-scoped `z-index` would only cover the join case while leaving standalone buttons worse off.

Sorry for the noise on the first version of this PR — the earlier comparison was simply not adversarial enough.

pdanpdan:
No problem, thank you for the research
Fixed play for otp-only change: https://play.tailwindcss.com/LEyZbb15eR

pdanpdan (approved):

PR #4660, opened 2026-08-06
docs(ember): fix the CSS import path in the install guide

### LINKED ISSUE #4658: docs: Outdated ember install instructions

https://daisyui.com/docs/install/ember/

### Describe the issue

a modern (current) ember app does not have a index.html file and hence the instructed snippet to import the CSS file is wrong

I'm not expert at all but I suppose you should instead modify the app.ts file adding a line like
> import './styles/app.css';

_No response_

JamBalaya56562:
closes #4658

## The bug

[The Ember guide](https://daisyui.com/docs/install/ember/) writes the Tailwind entry to `app/styles/app.css`:

```postcss:app/styles/app.css
@import "tailwindcss";
@plugin "daisyui";
```

…and then imports **`./app/styles.css`** — a path that does not exist (`styles/` is missing). Following the guide verbatim, `vite build` fails outright:

```
✗ Could not resolve "./app/styles.css" from index.html
```

So the project never builds. One character short of a directory, but it takes the whole guide down.

## About the report

@Polve's diagnosis was that a modern Ember app has no `index.html`, and that the CSS should be imported from `app.ts` instead. I scaffolded a fresh app to check, and **`index.html` does exist** — `@ember/app-blueprint@7.1.1` generates it at the project root (it's a Vite app, so it needs one):

```
create index.html
create app/styles/app.css
create vite.config.mjs
```

So the `index.html` step is fine as a location; only the path inside it is wrong. Fixing the path is enough — hence this one-line change rather than a restructure.

(For what it's worth, importing from `app/app.js` as suggested works too — I measured both, see below. I kept `index.html` because it's the smaller change and the existing wording stays accurate.)

## Verification

Scaffolded a real project and followed the guide step by step, changing only the import line:

```sh
npx ember-cli@latest init --blueprint @ember/app-blueprint
npm install tailwindcss@latest @tailwindcss/vite@latest daisyui@latest
# add tailwindcss() to vite.config.mjs, write app/styles/app.css, add the import to index.html
npm run build
```

Then served `dist/` and read the computed style of `<button class="btn btn-primary">` in Chromium:

| guide as written | with this fix |
| --- | --- |
| **build fails** — `./app/styles.css` cannot be resolved | builds in ~35s |
| — | CSS bundle **21.43 kB** (was 0.00 kB) |
| — | `display: inline-flex`, `background: oklch(0.45 0.24 277.023)`, `height: 40px`, `font-weight: 600` — daisyUI applied ✅ |

I also checked the two alternatives so the numbers are on the record:

| approach | CSS bundle | styled |
| --- | --- | --- |
| import removed entirely | 0.00 kB | ❌ plain UA button |
| `index.html` + `./app/styles/app.css` (**this PR**) | 21.43 kB | ✅ |
| `app/app.js` + `./styles/app.css` (the reporter's suggestion) | 21.43 kB | ✅ |

Removing the step is not an option: Embroider serves `app/styles/app.css` as a virtual module that never passes through `@tailwindcss/vite`, so without an import that routes it into Vite's module graph the file ships **unprocessed** — the built `@embroider/virtual/app.css` is literally the two source lines.

## One thing I noticed but did not change

With any of the working variants the browser logs a 404 for `/@embroider/virtual/tailwindcss`. Cause: Embroider *also* serves `app/styles/app.css` raw at `/@embroider/virtual/app.css`, so the browser tries to resolve the unprocessed `@import "tailwindcss"` as a URL. Harmless — the real, processed stylesheet loads from `/assets/main-*.css` — but it is a console error on every page load.

It goes away if the Tailwind entry lives outside `app/styles/`:

```postcss:app/tailwind.css
@import "tailwindcss";
@plugin "daisyui";
```
```diff:app/app.js
+ import './tailwind.css';
  import Application from '@ember/application';
```

Measured: same 21.43 kB, styles applied, **zero console errors and zero failed requests**. I left it out of this PR because it restructures the guide and adds a new translatable string across 27 locales — happy to send it as a follow-up if you want it.

## Translations

No translation changes: the sentence `"Import the CSS file in your index.html"` is unchanged, so its key stays in use. `bun run lang:prune` reports nothing to prune and `bun run lang:validate` passes.

### DIFF
--- a/packages/docs/src/routes/(routes)/docs/install/ember/+page.md
+++ b/packages/docs/src/routes/(routes)/docs/install/ember/+page.md
@@ -54,7 +54,7 @@ Import the CSS file in your index.html
 ```diff:index.html
 <head>
 +  <script type="module">
-+    import "./app/styles.css";
++    import "./app/styles/app.css";
 +  </script>
   <!-- the rest -->
 </head>


evoactivity:
updating the default ember @embroider/virtual/app.css link tag to 
```html
<link integrity="" rel="stylesheet" href="/app/styles/app.css">
```

is all that is required, no need to import through an ESM script tag.

Polve:
sorry for stating something wrong about index.html

The proposed modification from @evoactivity works, thanks

JamBalaya56562:
Thanks @evoactivity — you're right, and thanks for the correction. I rebuilt the app
from scratch and measured every variant. Your suggestion works and is strictly better
than what I pushed.

Two things turned up that I think are worth a decision before I change the patch, so
I'm posting the numbers rather than force-pushing blind.

### Results

| # | approach | build | processed CSS | styled | `/@embroider/virtual/tailwindcss` 404 | tests page styled |
|---|---|---|---|---|---|---|
| 0 | guide as written (`./app/styles.css`) | **fails**, exit 1 | — | — | — | — |
| 1 | remove the import step entirely | ok | **1 B** | ❌ | no | ❌ |
| 2 | this PR (`<script type="module">`) | ok | 21,434 B | ✅ | **yes** | ❌ |
| 3 | **your suggestion** (repoint the `<link>`) | ok | 21,434 B | ✅ | no | ❌ |
| 4 | `import './styles/app.css'` in `app/app.js` | ok | 21,434 B | ✅ | **yes** | ✅ |
| 5 | `app/tailwind.css` + import in `app/app.js` | ok | 21,434 B | ✅ | no | ✅ |

#2–#5 all give `display: inline-flex`, `background: oklch(0.45 0.24 277.023)`,
`height: 40px`, `font-weight: 600` on `<button class="btn btn-primary">`.

<details>
<summary>Environment and method</summary>

`ember-cli` 7.1.0 · `ember-source` 7.1.0 · `@embroider/vite` 1.7.9 ·
`@embroider/compat` 4.1.22 · `vite` 8.2.1 · `tailwindcss` 4.3.3 · `daisyui` 5.7.16 ·
Chrome 151.0.7922.71.

Fresh `npx ember-cli@latest init --blueprint @ember/app-blueprint`, guide followed
verbatim, only the CSS wiring changed per variant. Computed styles read from a real
browser; "processed CSS" is the size of the emitted `dist/assets/*.css`.

`favicon.ico` 404s in every variant including the baseline, so ignore that one.

One measurement note, since it tripped me up: the `/@embroider/virtual/tailwindcss`
404 only reproduces on a plain static server. `vite preview` answers it with
`200 text/html` via SPA fallback, so it can look clean when it isn't.

</details>

<details>
<summary>Why the step can't just be deleted (mechanism)</summary>

I originally wondered whether the import could simply be dropped (#1 above). It can't —
`/@embroider/virtual/app.css` is built by `@embroider/compat`, i.e. the classic
Broccoli pipeline:

```js
// @embroider/compat/dist/src/compat-app.js:390-401
let options = { outputPaths: { app: `/@embroider/virtual/app.css` }, ... };
let nestedInput = funnel(this.combinedStyles(addonTrees), { destDir: 'app/styles' });
let styles = this.preprocessors.preprocessCss(nestedInput, '/app/styles', '/assets', options);
```

`app/styles/app.css` never enters the Vite module graph, so `@tailwindcss/vite` can't
see it. `dist/@embroider/virtual/app.css` is 42 bytes — the two source lines verbatim,
which is why the browser then tries to resolve `@import "tailwindcss"` as a URL.

Your change fixes that in the most direct way, and the built `index.html` collapses to
a single processed stylesheet:

```html
<!-- #2 (this PR): unprocessed + processed, and the unprocessed one causes the 404 -->
<link integrity="" rel="stylesheet" href="/@embroider/virtual/app.css">
<link rel="stylesheet" crossorigin="" href="/assets/main-CTagfFYw.css">

<!-- #3 (yours): Vite rewrites it, one sheet -->
<link rel="stylesheet" crossorigin="" href="/assets/main-CTagfFYw.css">
```

</details>

### Two caveats I hit with #3

**1. v1 addon styles are dropped.** `/@embroider/virtual/app.css` merges more than
`app/styles` — it also picks up `_app_styles_` from classic addons. On a stock
blueprint that tree is empty (hence the 42 bytes), so no harm. But a reader applying
this to an existing app with v1 addons would silently lose their CSS.

<details>
<summary>Source</summary>

```js
// compat-app.js:370-388  combinedStyles()
let trees = addonTrees.map(tree => funnel(tree, { srcDir: '_app_styles_' }));
trees.push(appStyles);
```

</details>

**2. The tests page isn't covered.** `tests/index.html` has its own
`<link rel="stylesheet" href="/@embroider/virtual/app.css">`, which the change doesn't
touch. Listing the tests page's stylesheets and checking for `.btn` rules:

| approach | daisyUI in tests |
|---|---|
| #3 yours | ❌ no sheet contains `.btn` |
| #4 `app/app.js` import | ✅ 16 rules, `.btn` present |
| #5 separate entry | ✅ 16 rules, `.btn` present |

### Where that leaves it

- **#3** is the smallest correct fix, and on a stock blueprint it's completely right —
  build fixed, 404 gone, one line. Happy to switch to it.
- **#5** keeps the `/@embroider/virtual/app.css` link intact (so classic addon CSS
  still loads), covers tests, and ships no unprocessed CSS. Technically the most
  robust, but it restructures the guide — and it changes the wording
  "Import the CSS file in your index.html", which is a translated string across
  27 locales, so I'd rather not do that unilaterally.

I'll push whichever the maintainers prefer.

PR #4664, opened 2026-08-08
docs: fix $$ placement so responsive examples work with a configured prefix

### LINKED ISSUE #4663: docs: responsive examples break when a prefix is configured

https://daisyui.com/components/footer/#footer-with-logo-and-social-icons

### Describe the issue

set the docs prefix input to `d-` and the footer example shows:

`<footer class="d-footer d-sm:footer-horizontal bg-neutral text-neutral-content p-10">`

Image

there's no `d-sm:` variant, so copying that into a prefixed project leaves the footer vertical at every width. should be `sm:d-footer-horizontal`.

same in tooltip (`d-lg:tooltip`, `d-md:tooltip-right`, `d-md:tooltip-center`), label (the responsive floating-label rows), and drawer (`d-is-drawer-close:tooltip`, where the variant is fine but the class after it loses its prefix).

21 tokens across those 4 pages. identical output with the default empty prefix, which is why it's easy to miss.

can you assign this to me? i have the fix ready.

_No response_

sulimanbenhalim:
With a prefix configured, `$$sm:footer-horizontal` compiles to `d-sm:footer-horizontal` which doesn't exist, so the responsive modifier silently does nothing. Moved `$$` after the variant. The drawer examples also needed `$$` on the class after `is-drawer-close:`.

Output is byte identical with the default empty prefix.

close #4663

### DIFF
--- a/packages/docs/src/routes/(routes)/components/drawer/+page.md
+++ b/packages/docs/src/routes/(routes)/components/drawer/+page.md
@@ -283,7 +283,7 @@ You can check/uncheck the checkbox using JavaScript or by clicking the `label` t
       <ul class="$$menu w-full grow">
         <!-- List item -->
         <li>
-          <button class="$$is-drawer-close:tooltip $$is-drawer-close:tooltip-right" data-tip="Homepage">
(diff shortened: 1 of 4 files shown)


saadeghi:
Thank you 🙏

Issue #4349, opened 2025-12-16
docs(menu): fix invalid JSX class attributes


hawai1401:
### On which page do you see this issue?

https://daisyui.com/components/menu/#file-tree

### Describe the issue

Hello,

The JSX example for the tree menu mixes `class` and `className`.
This breaks React/JSX usage and is inconsistent with the rest of the snippet.

```jsx
<ul className="menu menu-xs bg-base-200 rounded-box max-w-xs w-full">
  <li>
    <a>
      <svg
        xmlns="http://www.w3.org/2000/svg"
        fill="none"
        viewBox="0 0 24 24"
        strokeWidth="1.5"
        stroke="currentColor"
        className="h-4 w-4">
        <path
          strokeLinecap="round"
          strokeLinejoin="round"
          d="M19.5 14.25v-2.625a3.375 3.375 0 00-3.375-3.375h-1.5A1.125 1.125 0 0113.5 7.125v-1.5a3.375 3.375 0 00-3.375-3.375H8.25m0 12.75h7.5m-7.5 3H12M10.5 2.25H5.625c-.621 0-1.125.504-1.125 1.125v17.25c0 .621.504 1.125 1.125 1.125h12.75c.621 0 1.125-.504 1.125-1.125V11.25a9 9 0 00-9-9z" />
      </svg>
      resume.pdf
    </a>
  </li>
  <li>
    <details open>
      <summary>
        <svg
          xmlns="http://www.w3.org/2000/svg"
          fill="none"
          viewBox="0 0 24 24"
          strokeWidth="1.5"
          stroke="currentColor"
          className="h-4 w-4">
          <path
            strokeLinecap="round"
            strokeLinejoin="round"
            d="M2.25 12.75V12A2.25 2.25 0 014.5 9.75h15A2.25 2.25 0 0121.75 12v.75m-8.69-6.44l-2.12-2.12a1.5 1.5 0 00-1.061-.44H4.5A2.25 2.25 0 002.25 6v12a2.25 2.25 0 002.25 2.25h15A2.25 2.25 0 0021.75 18V9a2.25 2.25 0 00-2.25-2.25h-5.379a1.5 1.5 0 01-1.06-.44z" />
        </svg>
        My Files
      </summary>
      <ul>
        <li>
          <a>
            <svg
              xmlns="http://www.w3.org/2000/svg"
              fill="none"
              viewBox="0 0 24 24"
              strokeWidth="1.5"
              stroke="currentColor"
[...]

PR #4665, opened 2026-08-08
fix(docs): convert class, for and checked to JSX at any indentation

sulimanbenhalim:
the JSX tab still shows `class=` on the File tree example from #4349. those svgs are indented 16 to 26 spaces but the attr-name map only enumerates up to 14 (`for` only 1), so they get missed. `strokeWidth` right next to it converts fine because it's a plain string key.

this matches the span with `\s*` instead so any depth works, and it drops the 16 enumerated entries.

i checked it against the live menu page markup: out of 46 code blocks only the previously missed `class` spans change, everything else is byte identical.

ref #4349


sulimanbenhalim:
also folded in `checked="checked"` -> `defaultChecked`, same map same cause. only converts at one leading space today, so these still miss: checkbox and toggle "with custom colors", pagination "using radio inputs".

PR #4666, opened 2026-08-08
feat: add neobrutalism theme pair

imxade:
https://play.tailwindcss.com/MTmgLlyhwo


pdanpdan:
Can you please add a play tailwind to showcase the themes?
I added an example:
~https://play.tailwindcss.com/fGq8E84peq~
https://play.tailwindcss.com/8CB8Ka8wNI - updated after theme change
Not bad from my pov

imxade:
@pdanpdan https://play.tailwindcss.com/MTmgLlyhwo

saadeghi:
Thanks for the PR.  
One of the core visual elements of a neobrutalism UI is that thick angled shadow (like the 4px 4px ones that you used in your demo).  
Currently we don't have a way to toggle such shadow style in a daisyUI theme (with a variable).  
I like this idea but without the shadows, it's hard to call it neobrutalism, and people need to add the shadows manually using utility classes, to get the expected look.

PR #4668, opened 2026-08-08
docs: add missing alt text to image examples so copied code matches the preview

### LINKED ISSUE #4667: docs: image examples are missing alt, and the existing alt does not describe the images
two related things about the `` examples.

first is a straight mismatch. every rendered preview has an `alt`, but 84 images in the copy-paste code blocks don't, so the code you copy differs from the example right above it. it's in avatar (17), carousel (29), mask (15), list (9), hover-gallery (8), stack (3), hero (2), stat (1). there's no case of both missing, which is why it looks like drift rather than a decision.

second is a question. the existing values look keyword-oriented rather than descriptive. `photo-1494253109108` is a blue sphere and half a lemon, and on the carousel page alone it's labelled `Carousel slider`, `Burger`, `Tailwind CSS slider`, `Pizza`, `Drink`, `Tailwind CSS component` and more, 11 different values for the same photo. same in hover-gallery, where all four hat photos (front, three-quarter, side and back of the same cap) share `Tailwind CSS image hover gallery`.

so which would you prefer:

1. sync the code blocks to the previews, keeping the current values exactly as they are
2. switch to descriptive alt (`Black cap with a daisy logo, front view` and so on), which means changing the previews too, around 203 images

i've got option 1 done and tested, PR coming right after this. i also went through all 33 unique images so i have descriptions ready if you want option 2 instead, or on top. my guess is the current values are deliberate for image search, so that part is your call rather than mine.

can you assign this to me?

sulimanbenhalim:
every preview in these examples has an `alt`, but 84 images in the copy-paste code blocks don't, so the code you copy differs from the example above it. this adds the missing ones by taking each image's `alt` from its own preview, so no existing value changes.

avatar 17, carousel 29, mask 15, list 9, hover-gallery 8, stack 3, hero 2, stat 1. everything except the added `alt` is byte identical, and no preview is touched. mask is the useful check: all 15 examples share one `src` but each code block picked up its own shape name.

on the second half of #4667, whether these should become descriptive (`Black cap with a daisy logo, front view` instead of `Tailwind CSS image hover gallery`): this PR deliberately doesn't touch that, since it would mean rewriting the previews too and the current values look chosen for image search. happy to follow up either way, i went through all 33 images so the descriptions are ready.

closes: #4667

### DIFF
--- a/packages/docs/src/routes/(routes)/components/avatar/+page.md
+++ b/packages/docs/src/routes/(routes)/components/avatar/+page.md
@@ -34,7 +34,7 @@ classnames:
 ```html
 <div class="$$avatar">
   <div class="w-24 rounded">
-    <img src="https://img.daisyui.com/images/profile/demo/batperson@192.webp" />
+    <img alt="Tailwind-CSS-Avatar-component" src="https://img.daisyui.com/images/profile/demo/batperson@192.webp" />
   </div>
 </div>
 ```
@@ -65,7 +65,7 @@ classnames:
 ```html
 <div class="$$avatar">
   <div class="w-32 rounded">
-    <img src="https://img.daisyui.com/images/profile/demo/superperson@192.webp" />
+    <img alt="Tailwind-CSS-Avatar-component" src="https://img.daisyui.com/images/profile/demo/superperson@192.webp" />
   </div>
 </div>
 <div class="$$avatar">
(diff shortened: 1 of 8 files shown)


saadeghi:
Thanks @sulimanbenhalim

PR #4670, opened 2026-08-09
docs: fix typos in docs pages, blog posts and skill files

sulimanbenhalim:
27 single word fixes, nothing reworded beyond the typo itself.

the biggest cluster is `MCP is a an API`, copy pasted across 16 pages in `docs/editor/*` and `docs/mcp/*`. the rest are `the the`, `is is`, `of of`, `to to`, `exisiting`, `indepecent`, `clicked of` (should be `clicked or`), and `A HTML from for filtering items` in a filter heading where `from` should be `form`.

four of them are `desc:` frontmatter (file-input, text-rotate, fab, and one blog post), so those show up in the class table and in page metadata.

two notes: the filter one is a heading so its anchor slug changes, and nothing links to it. and i left `a HTML` / `a LTR` / `a SVG` alone since that one is arguable.


saadeghi:
Thank you 🙏

PR #4672, opened 2026-08-09
docs: document the row-hover class on the table page

sulimanbenhalim:
`row-hover` is in `table.css` and in the generated class list, but it isn't in the class table on the docs page, so there's no way to find it.

it's also nicer than doing it by hand because it's zebra aware: `base-200` normally, `base-300` inside `table-zebra` so it still reads against the stripes.

worth noting the "Table with a row that highlights on hover" example uses `hover:bg-base-300` rather than this class. happy to switch it over in a follow up if you want, i left it alone since it changes how that example looks.

### DIFF
--- a/packages/docs/src/routes/(routes)/components/table/+page.md
+++ b/packages/docs/src/routes/(routes)/components/table/+page.md
@@ -14,6 +14,8 @@ classnames:
     desc: For <table> to make all the rows inside <thead> and <tfoot> sticky
   - class: table-pin-cols
     desc: For <table> to make all the <th> columns sticky
+  - class: row-hover
+    desc: For <tr> to highlight the row on hover
   size:
   - class: table-xs
     desc: Extra small size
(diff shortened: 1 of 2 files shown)


sulimanbenhalim:
pushed one more commit: `row-hover` was missing from `skills/daisyui/components/table.md` too, so it wasn't reaching `llms.txt` either. added it there as well, otherwise the docs page would list the class but the skill file still wouldn't.

saadeghi:
Thanks for the PR.  
`row-hover` must be removed (will be removed in next major version). Don't add it to the docs.  
It was introduced in an older version of daisyUI but it was quickly replaced by the better solution (`hover:bg-base-300`).  
The class named stayed in the source code for backward compatibility (but it won't be in user CSS if the class name is not being used)

`hover:bg-base-300` is a better way because user can choose the color.

PR #4674, opened 2026-08-10
fix(docs): convert the remaining html attribute names in the JSX tab

sulimanbenhalim:
the JSX tab leaves some attributes in their html form, so the code you copy logs `Invalid DOM property` warnings. one of them comes out mangled instead.

on `/components/dock/` the same svgs convert `stroke-width` and `stroke-linecap` but leave all 101 `stroke-miterlimit` alone. on `/components/otp/` `maxlength` converts and `autocomplete` and `inputmode` don't. on `/components/modal/` the same button converts `popovertarget` but turns `popovertargetaction` into `popoverTargetaction`, since the map compiles to one regex of alternatives and the shorter key matches first.

| attribute | should be | live |
|---|---|---|
| `stroke-miterlimit` | `strokeMiterlimit` | 101 kebab, 0 converted |
| `autocomplete` | `autoComplete` | 32 kebab, 0 converted |
| `inputmode` | `inputMode` | 32 kebab, 0 converted |
| `popovertargetaction` | `popoverTargetAction` | 4 as `popoverTargetaction` |

`popovertargetaction` has to sit before `popovertarget` or the shorter key still wins.

i checked every attribute in all 555 machine converted code blocks against react's `possibleStandardNames`, and with these four entries there are no mismatches left.

related: #4665 fixes a different gap in the same converter (indentation), they don't overlap.


saadeghi:
Thank you 🙏

PR #4680, opened 2026-08-18
fix(fab): don't force fab layout onto script/style/template children

sulimanbenhalim:
same rule as #4561 fixed for footer and #4679 for dock. `.fab > *` gives every direct child `flex`, so a build inlined `<script>` or `<style>` shows up next to the fab buttons.

example: https://play.tailwindcss.com/jxBrt3ObIH

measured: script `flex` 54x20, style 69x16 instead of `display: none`.

fix is `> *:not(:where(script, style, template))`. `:where()` keeps the guard at zero specificity, plain `:not(script, style, template)` would override child styles, it changed `.btn` gap here from 6px to 8px.


pdanpdan:
Can you please use `:where(:not(script, style, template))` instead of `:not(:where(script, style, template))` - it looks more readable - maybe also modify in footer

sulimanbenhalim:
done, plus dock and footer. footer was still on `:not(script, style, template)` which isn't zero specificity, so that one was an actual fix.

fwiw the two aren't interchangeable if the list ever grows, `:not()` takes a non-forgiving selector list:

`:not(:where(a, b, :future))` keeps working, `:where()` drops the unknown entry
`:where(:not(a, b, :future))` stops matching entirely, so the guard fails open

no difference with `script, style, template` though.

pdanpdan:
In footer can you also please drop the `&` in front of `& > *:where(:not(script, style, template))`

sulimanbenhalim:
going to keep `:not(:where(...))` on these, sorry for the flip flop.

the two only match the same while every entry in the list is supported. `:not()` takes a non-forgiving selector list, so if someone later adds an entry a browser doesn't know, `:where(:not(...))` stops matching entirely and the guard fails open with no error. `:not(:where(...))` drops the unknown entry and keeps working.

leaving the `&` in footer as is, it compiles identically so it's not worth the churn.

footer's specificity fix stays either way, that one was a real bug rather than a rename.

PR #4681, opened 2026-08-18
docs(tab): use button in the code blocks so copied tabs are keyboard reachable

sulimanbenhalim:
the previews on this page use `<button role="tab">` but every code block hands you `<a role="tab">`, and an `<a>` with no href isn't focusable. so the tabs you copy can't be reached by keyboard at all, and they never hit the `:focus-visible` ring `.tab` already defines.

measured with the page's own markup: `<a role="tab">` 0/3 tabs reachable, `<button role="tab">` 3/3.

this just makes the copied code match its preview, same idea as #4668. i checked every component docs page and tab is the only one where the preview and the code block use different tags (6 examples).

### DIFF
--- a/packages/docs/src/routes/(routes)/components/tab/+page.md
+++ b/packages/docs/src/routes/(routes)/components/tab/+page.md
@@ -57,9 +57,9 @@ classnames:
 
 ```html
 <div role="tablist" class="$$tabs">
-  <a role="tab" class="$$tab">Tab 1</a>
-  <a role="tab" class="$$tab $$tab-active">Tab 2</a>
-  <a role="tab" class="$$tab">Tab 3</a>
+  <button role="tab" class="$$tab">Tab 1</button>
+  <button role="tab" class="$$tab $$tab-active">Tab 2</button>
+  <button role="tab" class="$$tab">Tab 3</button>
 </div>
 ```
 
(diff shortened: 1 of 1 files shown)


pdanpdan:
I would go the other way around - I would change all to `<a href="#" ...` - I think this is the most used case.

saadeghi:
This is intentional.  
Usually these tabs are going to open links, so the code developer sees must be `<a>`.  
However we can't use links on the preview because it would either be invalid or it would open a new page or it would give lint errors.  In preview they are meant to be clicked and nothing should happen.  

And I wouldn't suggest `<a href="#">` because I don't want people to copy the code and forget to add hrefs. I want them to get an accessibility lint error on their app to tell them to add the `href`. Some accessibility linters won't show error on `href="#"` because it's valid code, but it's not how the developer is meant to use it, and most likely they want to add a real url for `href`.  

So  
- `<button>` for preview, so people can interact with it and we don't get accessibility errors.
- `<a>` for code so users gets accessibility lint warning and adds real link.

sulimanbenhalim:
i'd rather keep buttons here.

the page's own examples don't really support links being the common case: 13 tab examples, 6 radio and 6 anchor, and all 6 that build an actual tabbed panel with `tab-content` use radio inputs. the anchor ones are styling only and carry no `href`, so they aren't links today either.

`href="#"` also jumps the docs page to the top on every tab click, since the previews render inline.

and buttons are what the previews already render, which is the mismatch this pr is fixing.

PR #4694, opened 2026-08-21
fix(docs): give the bare popover attribute a value in the JSX tab

sulimanbenhalim:
a bare `popover` is `popover={true}` in JSX, and react types it as a string (`"" | "auto" | "hint" | "manual"`), so it warns and drops the attribute. the copied modal ends up as a plain div.

react 19.2.8, client render: `hasAttribute("popover")` is false and `showPopover()` throws `NotSupportedError`. `modal.css` keys off `[popover]` and `:popover-open`, so it never shows either.

36 of them, on megamenu, modal and calendar. the hand-written jsx example already writes `popover="auto"`.


saadeghi:
Thanks

PR #4700, opened 2026-08-24
fix(toast): keep toast-center horizontally centered in RTL

sulimanbenhalim:
toast-center anchors with the logical start-1/2 end-1/2 but the --toast-x: -50% translate is physical, so in RTL the toast lands a full width left of center. flipping the translate in RTL fixes it, same pattern modal-start/end already use.

bug + fix: https://play.tailwindcss.com/AVbg7dPxAy


AndreasTschuden (approved):
Super

saadeghi:
Thanks @sulimanbenhalim 🌼

PR #4701, opened 2026-08-24
fix(label): keep the floating label anchored to its corner in RTL

sulimanbenhalim:
the floated label scales to 0.75 and shifts translate: -12.5% so its start edge stays in place. the translate is physical though, so in RTL the label drifts away from the corner by 25% of its width. flipped it for RTL, and the matching ::placeholder rule too so the focus transition moves the same way.

bug + fix: https://play.tailwindcss.com/GwlGlCzIms


saadeghi:
Thank you

PR #4702, opened 2026-08-24
feat(swap): swap between any number of items with swap-multi

sulimanbenhalim:
"swap with multiple items" from the roadmap. put a radio input before each item, the checked one shows its item and the input after it sits on top, so every click shows the next item and it wraps around. pure css, arrow keys work too since it is a radio group, classic 2-state swap untouched.

demo: https://play.tailwindcss.com/LmMO71fCSL

### DIFF
--- a/packages/daisyui/src/components/swap.css
+++ b/packages/daisyui/src/components/swap.css
@@ -70,6 +70,36 @@
   }
 }
 
+.swap-multi {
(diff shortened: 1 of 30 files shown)


sulimanbenhalim:
added a focus ring on swap-multi, the invisible radios were hiding the browser's own, so keyboard users had no indicator.

PR #4703, opened 2026-08-24
feat(marquee): add marquee component for infinite scrolling content

sulimanbenhalim:
marquee is on the roadmap under Future. pure css: two copies of the content inside the container (second one aria-hidden), animation on the copies. modifiers: marquee-reverse, marquee-vertical, marquee-pause-hover, speed via --marquee-duration. flips direction automatically in RTL and pauses under prefers-reduced-motion. docs page, skill file and translation keys included.

demo on the published plugin with the proposed css: https://play.tailwindcss.com/8osowxZoNb

### DIFF
--- a/packages/daisyui/build.js
+++ b/packages/daisyui/build.js
@@ -107,6 +107,7 @@ async function generateFiles() {
           "validator",
           "hover3d",
           "textrotate",
+          "marquee",
         ],
         layer: "utilities",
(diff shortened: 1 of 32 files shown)


sulimanbenhalim:
marquee across 12 themes: https://play.tailwindcss.com/5nTnV2T8lv

PR #4704, opened 2026-08-25
feat(ribbon): add ribbon component for corner labels on cards

sulimanbenhalim:
adds the ribbon component from the roadmap: a corner label for cards like "NEW" or "SALE", with a folded edge. pure css, flips with RTL, and the fold darkens whatever background color the ribbon has so all color modifiers work.

example: https://play.tailwindcss.com/Gj3QinkjK9


sulimanbenhalim:
ribbon across 12 themes: https://play.tailwindcss.com/BOVPPdx1oB

PR #4706, opened 2026-08-25
docs: update the WindPress link to its new home

sulimanbenhalim:
wind.press expired in july (whois says pendingDelete, dns is gone) so the wordpress guide's plugin link was dead. windpress moved to https://windpress.jooo.si/ (same site, same author), so the link points there now, in the guide and in all 27 translation files.


saadeghi (changes requested):
Thanks for the PR. Can we link to their GitHub page instead of domain? https://github.com/wind-press/windpress

sulimanbenhalim:
done, swapped the link for the github repo

PR #4712, opened 2026-08-31
fix(skeleton): mirror the shimmer direction in RTL

sulimanbenhalim:
the skeleton shimmer animates background-position, which is physical, so in RTL it still sweeps left to right while everything else reads right to left. one line plays the same animation reversed for :dir(rtl). reduced motion is untouched.

bug and fix: https://play.tailwindcss.com/4yhPZYykkR


saadeghi:
Thanks

PR #4717, opened 2026-09-01
fix(megamenu): lock page scroll while the fullscreen menu is open

sulimanbenhalim:
the fullscreen megamenu state has a backdrop but the page behind it keeps scrolling. modal and drawer both set --page-scroll-lock for this, megamenu just missed it.

bug: https://play.tailwindcss.com/S4VXxIOFnB
with the fix: https://play.tailwindcss.com/mCeZbLKGUB

### DIFF
--- a/packages/daisyui/src/components/megamenu.css
+++ b/packages/daisyui/src/components/megamenu.css
@@ -179,6 +179,9 @@
       @starting-style {
         @apply opacity-0;
       }
+      :root:has(&) {
+        --page-scroll-lock: ;
+      }
       [popovertarget] {
         @apply pointer-events-none;
         font-size: 0.875rem;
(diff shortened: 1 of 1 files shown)


pdanpdan:
The idea is good in theory, but only for the case where the megamenu is on top.
The megamenu popover tries to place itself below the activator button, so you might end up with it outside the screen and not-scrollable in screen.

Current: https://play.tailwindcss.com/0vcR2TqrIk
Proposed: https://play.tailwindcss.com/ylWWGYCThx

pdanpdan:
Now there is a nasty effect on close.

pdanpdan:
I found a solution to use available height in https://github.com/saadeghi/daisyui/pull/4718
So now the scroll-lock can be safely applied: https://play.tailwindcss.com/apCvCn0BaZ

sulimanbenhalim:
verified the two together: scrolled invoker, resize while the menu is open, wheel at the menu scroll end, rtl. the lock holds everywhere and the branches merge clean independently

pdanpdan:
You should not remove the 2 position area.

PR #4719, opened 2026-09-01
fix(join): keep the border overlap for join items inside a wrapper

sulimanbenhalim:
a join item inside a wrapper (like a tooltip) misses the negative margin, so each seam shows a double border. the radius vars already handle wrapped items, now the margin does too.

example: https://play.tailwindcss.com/1xhtLZD53I

seen in #4709


pdanpdan:
LGTM
https://play.tailwindcss.com/1Mh32enXQz

saadeghi:
Thanks 🌼

PR #4721, opened 2026-09-02
feat(megamenu): style plain links as items

sulimanbenhalim:
plain links between the popover buttons get the item style, the same sliding highlight, and the menu item look in vertical mode. also resets the leftover closed state transform on vertical panels.

example: https://play.tailwindcss.com/4r8gyMqyzV

as discussed in #4675


saadeghi (changes requested):
Thanks for the PR.   
Other than `<a>`, `<button>` should be allowed as well.  
So people can use a button to open a modal or something (for example a button that opens a login modal)

sulimanbenhalim:
done, buttons work too now, same style and highlight: https://play.tailwindcss.com/sNau8WCquQ (the Login button opens a modal)

saadeghi (changes requested):
Thanks 🫡  
One issue: if the modal button is [before](https://play.tailwindcss.com/DzPTy6w0nY) or [between](https://play.tailwindcss.com/PfMA9SFmxj) the popover buttons, the `megamenu-active` position would be wrong. (the popovers are not highlighted correctly when open)

sulimanbenhalim:
fixed, counted with nth-child(n of ...) now. before/after: https://play.tailwindcss.com/oGW4n8JhsY

sulimanbenhalim:
side note, the unstyled menu in the bot's After play is not from this PR: the bot excludes megamenu and a single exclude value also drops menu (substring match). same in the After of #4717 and in yesterday's After here. fix in #4726

PR #4724, opened 2026-09-02
fix(breadcrumbs): focus ring is clipped on the first and last item

sulimanbenhalim:
the list scrolls with no inline padding, so the focus ring of the first crumb (and the last one after scrolling) is cut off. 4px padding on the list gives it room, `min-width: max-content` keeps that padding after the last item. text moves 4px inward.

example: https://play.tailwindcss.com/6DmfYSbFaI

### DIFF
--- a/packages/daisyui/src/components/breadcrumbs.css
+++ b/packages/daisyui/src/components/breadcrumbs.css
@@ -1,11 +1,11 @@
 .breadcrumbs {
   @layer daisyui.l1.l2.l3 {
-    @apply max-w-full overflow-x-auto py-2;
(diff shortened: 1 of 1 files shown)


saadeghi (changes requested):
If we're adding padding, we should add a negative margin (`-ms-1`) so it will align with the rest of the content:
https://play.tailwindcss.com/gisT9lAwde?file=css

Also the crop seems to be happening only at the start. Maybe `ps-1` would be better than `px-1`?

sulimanbenhalim:
done, -ms-1 on the container and ps-1 on the list

PR #4726, opened 2026-09-03
fix(plugin): a single exclude value also drops components whose name it contains

sulimanbenhalim:
`exclude: megamenu;` also removes menu, `exclude: table;` also removes tab, same for status/stat, fileinput/input and radialprogress/progress. a single value reaches the plugin as a string and `exclude.includes(name)` becomes a substring check. comma lists come in as arrays and are fine. the handler now turns include and exclude into a list first.

single value: https://play.tailwindcss.com/DsHa2Oa2MH vs list: https://play.tailwindcss.com/lOXMrrTl1m (same menu, only the exclude form differs)

this is also why the preview bot's After playgrounds for megamenu PRs show an unstyled menu, the bot excludes megamenu.

### DIFF
--- a/packages/daisyui/functions/pluginOptionsHandler.js
+++ b/packages/daisyui/functions/pluginOptionsHandler.js
@@ -12,6 +12,16 @@ export const pluginOptionsHandler = (() => {
       prefix = "",
     } = options || {}
 
+    const toList = (value) =>
+      Array.isArray(value)
+        ? value
+        : String(value)
+            .split(",")
+            .map((item) => item.trim())
+            .filter(Boolean)
+    const includeList = include === undefined ? undefined : toList(include)
+    const excludeList = exclude === undefined ? undefined : toList(exclude)
+
     if (logs !== false && firstRun) {
       console.log(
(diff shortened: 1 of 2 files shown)


saadeghi (changes requested):
thanks for the PR.  

This PR is also changing the way how an empty include works.

Currently an empty include means include everything.  

As documented here https://daisyui.com/docs/config/ That's the expected behavior.  

Changing this behavior right now may break some apps so I would suggest keeping the same behavior for an empty include.

So if include is empty, it means it should include everything. (We can adjust the behavior of an empty include the next major version, but not now)

sulimanbenhalim:
right, thanks. a blank `include: ;` reaches the plugin as `0` and the list conversion turned it into `[]`, so nothing was included. pushed a follow up: include is untouched now (blank still means everything), only a single exclude value gets turned into a list. verified on vite, postcss and the cli with tailwind 4.0.0, 4.1.18 and 4.3.3 using the default config block from the docs, same output as 5.7.34 except the exclude fix.

PR #4729, opened 2026-09-03
fix(fab): labels render at 16px instead of 14px when daisyui has a prefix

sulimanbenhalim:
with `prefix: "d-"` the labels next to the fab buttons come out at 16px instead of the 14px you get without a prefix.

the fab is the only component that uses `text-sm` through @apply. that compiles to tailwind's `--text-sm` variable and the prefix step renames it to `--d-text-sm`, which doesn't exist. every other component just writes the rem values, so this does the same.

repro with prefix d-, the second fab has the fix: https://play.tailwindcss.com/fbfm9rqI5W


saadeghi:
Thank you

PR #4730, opened 2026-09-03
fix(otp): digits after the first sit one border width right of the box centre

sulimanbenhalim:
digits after the first sit 1px right of their box centre, 2px on themes with a 2px border. the input's padding-inline-start includes var(--border), so the whole text run is offset by the border and only the first box compensates, with a hard-coded left: 1px.

that 1px is really the caret: an empty field-sizing input is 1px wide, and the :after marker that follows the input needs it only while nothing is typed. this drops the border term from the padding and the :after margin, sets the first box to left: 0, and moves 1px of the inset from padding to text-indent (counts for typed text, not for the empty input). checked in chrome and firefox.

play, lower two rows restate the fix: https://play.tailwindcss.com/5aiPsIjP9r

### DIFF
--- a/packages/daisyui/src/components/otp.css
+++ b/packages/daisyui/src/components/otp.css
@@ -21,7 +21,8 @@
     > input {
       @apply pointer-events-none start-0 z-1 m-0 appearance-none border-0 bg-transparent p-0 outline-0;
       field-sizing: content;
-      padding-inline-start: calc(var(--otp-ch) * 0.5 + var(--border));
+      padding-inline-start: calc(var(--otp-ch) * 0.5 - 1px);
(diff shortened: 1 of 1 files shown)


pdanpdan:
I think that after these changes on the docs site on ios safari:
- the first digit is one 1px to the left (this is not so important)
- the digits move 1px to the left each time the field is focused (vs when not focused)
It does reproduce in the codepen on top also

sulimanbenhalim:
this happens before this pr too, same 1px move on focus with the 5.7.32 cdn file on an iphone: https://play.tailwindcss.com/ybekC4hIuE

PR #4751, opened 2026-09-11
fix(reset): keep elements with the hidden attribute hidden in daisyui.css

sulimanbenhalim:
with the standalone `daisyui.css` (no tailwind, so no preflight) a `<button class="btn" hidden>` still renders, same for badge, card, alert, menu, input: the component's `display` beats the browser's `[hidden]` rule. `reset.css` trimmed preflight but dropped its `[hidden]` rule, this adds it back as is (preflight uses `!important` there too, it has to beat any `display`). plugin users are not affected, they get preflight.

example: https://play.tailwindcss.com/PhJiCBxS79

### DIFF
--- a/packages/daisyui/build.js
+++ b/packages/daisyui/build.js
@@ -88,7 +88,13 @@ async function generateFiles() {
 
     !isDev && generateThemeFiles({ srcDir: "src/themes", distDir: "theme" }),
 
-    !isDev && generateRawStyles({ srcDir: "../src/base", distDir: "../base", layer: "base" }),
+    !isDev &&
+      generateRawStyles({
+        srcDir: "../src/base",
+        distDir: "../base",
+        layer: "base",
+        unlayered: ["hidden"],
+      }),
 
     !isDev &&
       generateRawStyles({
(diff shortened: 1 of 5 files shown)


pdanpdan:
Watch out for hidden until-found

sulimanbenhalim:
already excluded
its preflight's `[hidden]:where(:not([hidden="until-found"]))` as is and i checked that until-found stays findable

saadeghi (changes requested):
Is there a way to do it without `!important`?

sulimanbenhalim:
tried, no. reset is packed in `@layer base` and components sit in `utilities.daisyui.*`, layer order wins so a plain rule in base never applies (a 4x specificity selector loses too). it only works unlayered or in the `utilities` layer, which needs a packCss change. preflight keeps `!important` here for the same reason

sulimanbenhalim:
found a way. the rule now lives in its own `src/base/hidden.css` and `generateRawStyles` skips the `@layer base` wrap for it, so it ships unlayered in `daisyui.css` and wins on cascade order alone, no `!important`. plugin build untouched (excluded like reset), chunks.css picks it up too

Issue #4054, opened 2025-08-14
bug: Tab active not working with `aria-current`


filips123:
### Reproduction URL (Required)

https://play.tailwindcss.com/nvz7BvXTmw

### What version of daisyUI are you using?

v5.0.50

### Which browsers are you seeing the problem on?

All browsers

### Describe your issue

Using `aria-[current='page']` modifier doesn't work when applying it to `tab-active` and `menu-active` classes:

```html
<div role="tablist" class="tabs">
  <a role="tab" class="tab aria-[current='page']:tab-active">Tab 1</a>
  <a role="tab" class="tab aria-[current='page']:tab-active" aria-current="page">Tab 2</a>
  <a role="tab" class="tab aria-[current='page']:tab-active">Tab 3</a>
</div>
```

<img width="294" height="66" alt="Image" src="https://github.com/user-attachments/assets/78ff5600-537f-4732-9908-84fea6cc5048" />

This is the same issue as #3652, but the fix was incomplete (only for `btn-active`). Maybe there are also other similar classes that won't work.

PR #4752, opened 2026-09-11
fix(menu): style items with aria-current like menu-active

sulimanbenhalim:
a menu item with `aria-current="page"` looks like every other item, while the same attribute already gets the active style on tabs (#4054 asked for both). react router's NavLink and vue router's RouterLink put `aria-current="page"` on the active link by themselves, so a menu built from them has no active item. this adds `[aria-current="true"]` and `[aria-current="page"]` next to `.menu-active`, hover excluded the same way.

example: https://play.tailwindcss.com/HPnjNS50Z3


pdanpdan:
I would go with `[aria-current]:not([aria-current="false"], [aria-current=""])`

sulimanbenhalim:
makes sense pushed it

goransh-walia:
Nice, straightforward implementation.

PR #4753, opened 2026-09-11
fix(dock): show the active indicator on items with aria-current

sulimanbenhalim:
a dock item with `aria-current="page"` gets no indicator, only `dock-active` does, and that is the attribute react router's NavLink and vue router's RouterLink set on the active link. this treats `aria-current="true"` and `"page"` the same way `dock-active` is, sizes included, like tab does.

example: https://play.tailwindcss.com/ZvegZ2Swip

### DIFF
--- a/packages/daisyui/src/components/dock.css
+++ b/packages/daisyui/src/components/dock.css
@@ -35,6 +35,11 @@
       }
     }
   }
+  @layer daisyui.l1.l2 {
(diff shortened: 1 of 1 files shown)


pdanpdan:
I would go with `[aria-current]:not([aria-current="false"], [aria-current=""])`

sulimanbenhalim:
makes sense pushed it

PR #4759, opened 2026-09-15
fix(tooltip): tail notch on tooltip-left and tooltip-right with a pill bubble

### LINKED ISSUE #4757: bug: FAB tooltip arrow

https://daisyui.com/components/fab/

_No response_

All browsers

### Describe your issue

when the arrow in on the side and the button has large radius, the surface is not smooth. you can vividly see the bump on the curve.

Image

sulimanbenhalim:
on themes with a big field radius the side tooltips show a little step above and below the tail, the tail base is flat and pokes out where the pill curves in. moved the side tails 1px further in so it stays hidden.

example: https://play.tailwindcss.com/gQncpVXvKK

close #4757


saadeghi (changes requested):
Thanks but it won't look good on other themes.  

Before/after:  

Image  
See how the soft edges are sharp.

Maybe instead of hardcoding 1px or 2px use something like
```css
clamp( 1px, calc(var(--radius-field) / 1rem * 1px), 2px )
```
so it will be based on the radius?

saadeghi:
Also, I think tooltip-start and tooltip-end should be adjusted too.

Screenshot 2026-09-17 at 5 20 50 PM

Screenshot 2026-09-17 at 5 22 30 PM

sulimanbenhalim:
pushed a follow up, the overlap and the start/end offsets follow the radius now (divided by a number, firefox computes the `/ 1rem * 1px` form to 0). still not perfect on every theme, before and after per radius tier: https://play.tailwindcss.com/aghqAeTNcv

PR #4763, opened 2026-09-16
fix(diff): tap to expand each side on Firefox for Android

### LINKED ISSUE #4762: Diff component not working on Firefox Android
> Firefox mobile build 155.0.1 (Build #2016182535), 5fdfd0092780e85643e2cddc0e1b590c8b9ef860
> GV: 155.0.1-20260903215306
> AS: 155.0
> OS: Android 17
> 
> even on the documentation page you cannot drag the diff changer. It's stuck in the middle
>  

 _Originally posted by @dsoh in [#4733](https://github.com/saadeghi/daisyui/issues/4733#issuecomment-5691187749)_

sulimanbenhalim:
firefox for android parses `resize` but ignores it (mdn: https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/resize#browser_compatibility, bug 1776834), so the resizer can't be dragged, same as ios safari. this gives gecko on touch the tap fallback the ios block already has, tap a side to expand it, and hides the drag thumb there. desktop firefox is untouched.

example: https://play.tailwindcss.com/ZQ5blKvil7

close #4762


saadeghi:
Thanks 👏

PR #4764, opened 2026-09-17
docs: keep english-only blocks left to right on rtl languages

sulimanbenhalim:
with an rtl language (arabic, persian, hebrew, urdu) the english-only parts of the site sit in an rtl page, so the bidi algorithm throws their trailing punctuation to the front and the changelog renders right aligned with mirrored bullets. `dir="ltr"` on the four of them: the top banner (`!daisyUI MCP v1.6: Generate unique UIs now`), the carbon ad, the testimonial quotes on the homepage (`!It's great`) and the changelog body, like the changelog menu and the search box already have.


saadeghi:
Thank you 🫡

Issue #4465, opened 2026-03-14
bug: Range slider severely laggy on Chromium due to 100rem box-shadow track fill


jmynes:
## Description

The `range` component is extremely laggy when dragging on Chromium-based browsers (Chrome, Edge, Brave). Firefox handles it fine. This is reproducible on the official DaisyUI demo page: https://daisyui.com/components/range/

## Root Cause

The filled track effect uses a massive `box-shadow` on the thumb pseudo-element:

```css
.range::-webkit-slider-thumb {
  box-shadow:
    /* ... other shadows ... */
    calc((var(--range-dir, 1) * -100rem) - (var(--range-dir, 1) * var(--range-thumb-size) / 2)) 0 0 calc(100rem * var(--range-fill));
}
```

The `100rem` spread radius creates a huge paint area that Chromium's renderer has to repaint on every thumb position change. Firefox's renderer handles this more efficiently.

## Steps to Reproduce

1. Open https://daisyui.com/components/range/ in Chrome/Chromium
2. Drag any range slider rapidly left and right
3. Observe significant input lag and janky thumb movement

Compare with Firefox where the same slider is buttery smooth.

## Environment

- DaisyUI 5.x (reproducible on latest)
- Chrome 130+ / any Chromium-based browser
- Any OS (tested on Linux, reported by users on Windows)

## Suggested Fix

Replace the box-shadow track fill with a CSS `linear-gradient` on the track, or use a `clip-path` approach. For example:

```css
.range::-webkit-slider-runnable-track {
  background: linear-gradient(
    to right,
    currentColor var(--fill-percent),
    var(--range-bg) var(--fill-percent)
  );
}
```

This requires calculating the fill percentage (either via a CSS custom property set from JS, or using a container-based approach), but eliminates the expensive box-shadow repaint entirely.

MUI's slider avoids this problem by not using `<input type="range">` at all — it uses a div-based slider with `role="slider"` and JS-driven positioning.

pdanpdan:
Hello.
Are you sure you are looking at the current daisyUI version (just asking because 100rem was replaced with 100cqw). I don't think it would improve anything for your case, but just to start from the same base.

The components should work without JS.

I have created a play - you can test various solutions here and post if you find something interesting:
https://play.tailwindcss.com/zRQn4OI9vn?file=css

[...]

PR #4771, opened 2026-09-19
fix(range): give the fill a query container so cqw is track-relative

SurefireStudios:
`--range-fill-x` and `--range-fill-spread` are sized in `cqw`, but `.range` declares no `container-type`, so those units resolve against the viewport instead of the track. The thumb's fill shadow is painted at full viewport width — 1200px in a 1200px window, and wider on a large monitor — which is the oversized paint area behind the Chromium drag lag in #4465. That issue blames the old `100rem`, but `100rem` is 1600px and today's `100cqw` can exceed it.

Adding `container-type: inline-size` scopes the units to the input, so the spread becomes the track's own width. Measured in Chromium at a 1200px viewport: `10cqw` on a 320px element resolves to **120px without** a container and **32px with** one, and `container-type` on the element does scope the units for its `::-webkit-slider-thumb`. Rendering is unchanged — two sliders side by side, one patched, are pixel-identical. `bun run build` passes and the built `.range` rule carries the declaration.


saadeghi:
Thanks @SurefireStudios

